PR #2, opened 2021-11-06
first crack at a pipeline to download flow data and compute some metrics

slevin75:
This pipeline downloads daily flow data, as well as peak data and some site info needed for EflowStats, runs all the HIT metrics for each site and then compiles it back into one file.  Theres a lot of comments in the code where future functions will probably be needed or just notes to myself.  Also - I wasn't quite sure what to save as a file and what to keep as an object target, so other than the initial daily streamflow download, they are all objects but I can change them as needed. I'm sure you will find a lot of other stuff in there, I was just pretty psyched that it ran!

### DIFF
--- a/1_fetch/src/calc_HIT.R
+++ b/1_fetch/src/calc_HIT.R
@@ -1,0 +1,9 @@
+
+
+
+calc_HITmetrics<-function(data,yearType,drainArea,floodThreshold){
+  
+  out_data<-calc_allHIT(x=data,yearType=yearType,drainArea=drainArea,floodThreshold=floodThreshold)
+  out_data$site_num<-unique(data$site_no)
+  return(out_data)
+}
--- a/1_fetch/src/calc_HIT.R
+++ b/1_fetch/src/calc_HIT.R
@@ -2,7 +2,7 @@
 
 
 calc_HITmetrics<-function(data,yearType,drainArea,floodThreshold){
-  
+  suppressWarnings(calc_allHIT(x=data,yearType=yearType,drainArea=drainArea,floodThreshold=floodThreshold))
   out_data<-calc_allHIT(x=data,yearType=yearType,drainArea=drainArea,floodThreshold=floodThreshold)
   out_data$site_num<-unique(data$site_no)
   return(out_data)
--- a/1_fetch/src/calc_HIT.R
+++ b/1_fetch/src/calc_HIT.R
@@ -2,8 +2,10 @@
 
 
 calc_HITmetrics<-function(data,yearType,drainArea,floodThreshold){
-  suppressWarnings(calc_allHIT(x=data,yearType=yearType,drainArea=drainArea,floodThreshold=floodThreshold))
-  out_data<-calc_allHIT(x=data,yearType=yearType,drainArea=drainArea,floodThreshold=floodThreshold)
+  out_data<- suppressWarnings(calc_allHIT(x=data,yearType=yearType,
+                                          drainArea=drainArea,
+                                          floodThreshold=floodThreshold))
+ 
   out_data$site_num<-unique(data$site_no)
   return(out_data)
 }
--- a/1_fetch/src/get_nwis_data.R
+++ b/1_fetch/src/get_nwis_data.R
@@ -2,7 +2,9 @@
 get_nwis_daily_data<-function(site_num,outdir,parameterCd,startDate, endDate){
    ##read daily NWIS data, and save to csv file.
   data_out<-readNWISdv(site_num, parameterCd, startDate, endDate)
-  
+  ##renaming column names for discharge and discharge_cd
+  names(data_out)[grep(pattern="X_.*\\cd",x=names(data_out))]<-"discharge_cd"
+  names(data_out)[grep(pattern="X_",x=names(data_out))]<-"discharge"
   filepath<-file.path(outdir,paste0(site_num, "_dv.csv"))
   write_csv(data_out, file=filepath)
   return(filepath)
@@ -26,12 +28,16 @@ screen_daily_data<-function(filename,yearType){
   ##screen for years with missing data
   data<-read.csv(filename)
   data$Date <- as.Date(data$Date)
+  ###prior to screening, remove any provisional data - this will be counted as 'no data'
+  prov_data<- grep("P",data$discharge_cd)
+  if(length(prov_data)>0){data<-data[-prov_data,]}
+  
   if(yearType=="water"){
     water_year_start<-10
   }else{
       water_year_start <- 1
     }
-  missing_data<-screen_flow_data(data.frame(site_no=data$site_no, Date=data$Date,Value=data$X_00060_00003),water_year_start=water_year_start)
+  missing_data<-screen_flow_data(data.frame(site_no=data$site_no, Date=data$Date,Value=data$discharge),water_year_start=water_year_start)
   return(missing_data)
 }
 
@@ -44,7 +50,7 @@ clean_daily_data<-function(filename,missing_data, yearType ){
   data$Date <- as.Date(data$Date)
   
   ##remove any missing flow data
-  data<-data[which(!is.na(data$X_00060_00003)),]
+  data<-data[which(!is.na(data$discharge)),]
   data<-addWaterYear(data)
   
   ##remove all data from years with data gaps
@@ -56,7 +62,7 @@ clean_daily_data<-function(filename,missing_data, yearType ){
   }
   
   ###run EflowStats validation to produce clean, ready to process data
-  clean_data<-validate_data(data_sc[,c("Date","X_00060_00003")],yearType=yearType)
+  clean_data<-validate_data(data_sc[,c("Date","discharge")],yearType=yearType)
   clean_data$site_no<-unique(data_sc$site_no)
   return(clean_data)
 }  #end clean_daily_data function
--- a/1_fetch/src/get_nwis_data.R
+++ b/1_fetch/src/get_nwis_data.R
@@ -1,0 +1,79 @@
+
+get_nwis_daily_data<-function(site_num,outdir,parameterCd,startDate, endDate){
+   ##read daily NWIS data, and save to csv file.
+  data_out<-readNWISdv(site_num, parameterCd, startDate, endDate)
+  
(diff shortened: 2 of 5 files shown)


_targets.R
```diff
@@ -0,0 +1,83 @@
+library(targets)
+library(tarchetypes)
+library(dataRetrieval)
+library(EflowStats)
+library(tidyverse)
+library(lubridate)
+library(fasstr)
```
jds485, line 7: Are all of these libraries needed in the main script? If they're only needed in some targets, you can specify them in a vector in `tar_option_set(packages = c(<package name>))`

slevin75: yes, most of those could probably go in the tar_option line.

1_fetch/src/get_nwis_data.R
```diff
@@ -0,0 +1,79 @@
+
+get_nwis_daily_data<-function(site_num,outdir,parameterCd,startDate, endDate){
+   ##read daily NWIS data, and save to csv file.
+  data_out<-readNWISdv(site_num, parameterCd, startDate, endDate)
+
```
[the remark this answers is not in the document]
jds485: Comment also applies to the peak flow data

_targets.R
```diff
@@ -0,0 +1,83 @@
...
+             screen_daily_data(p1_daily_flow,yearType)),
+  
+  ##clean data and reformat for EflowStats 
+  tar_target(p1_clean_daily_flow,
+             clean_daily_data(p1_daily_flow,p1_screen_daily_flow,yearType="water")),
+  
+  ##get drainage area (needed for EflowStats - can change this later if DA is part of the
+  ## basin characteristics file)
+  tar_target(p1_drainage_area,
+             readNWISsite(siteNumber=sites)$drain_area_va),
```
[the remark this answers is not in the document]
jds485: I see that there is a contributing drainage area option as well. It sounds like contributing area would always be smaller than drainage area (e.g., due to ponding of water or transfers of water). Should we opt to use contributing area instead?

1_fetch/src/get_nwis_data.R
```diff
@@ -0,0 +1,79 @@
...
+  ##Removing any NA dates or peak_va values - dates are returned as NA if day of occurance is not known
+  ## and peak_va is sometimes returned as NA if the gage height is known but not the discharge.
+  data_out<-data_out[which(!is.na(data_out$peak_dt)),]
+  data_out<-data_out[which(!is.na(data_out$peak_va)),]
+  return(data_out)
+  
+}
+
+
+screen_daily_data<-function(filename,yearType){
```
jds485, line 25: Could add a few lines to remove all provisional data. I'd think that stopping in 2020 will already get most of those, though.

1_fetch/src/calc_HIT.R
```diff
@@ -0,0 +1,9 @@
+
+
+
+calc_HITmetrics<-function(data,yearType,drainArea,floodThreshold){
+  
+  out_data<-calc_allHIT(x=data,yearType=yearType,drainArea=drainArea,floodThreshold=floodThreshold)
```
jds485, line 6: Let's suppress warnings for the `calc_allHIT` function because all of the warnings are "future warnings" for the developers of the `calc_allHIT` function.
`suppressWarnings(calc_allHIT(x=data,yearType=yearType,drainArea=drainArea,floodThreshold=floodThreshold))`

_targets.R
```diff
@@ -0,0 +1,83 @@
...
+
+
+##targets
+list(
+  p1_flow_metrics,
+  
+  ##combine flow metrics from all sites into one dataframe - not sure if that is 
+  ##the format we want for this or if it would be better as a list?
+  tar_combine(p1_combined_HIT_metrics,
+              p1_flow_metrics[[6]],
```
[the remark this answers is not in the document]
jds485: Also, this is currently grabbing the flood thresholds instead of the HIT metrics

1_fetch/src/get_nwis_data.R
```diff
@@ -0,0 +1,79 @@
...
+
+
+calc_water_year<-function(dates){
+  water_year <- ifelse(month(dates) > 9, year(dates)+1,year(dates) )
+  return(water_year)
+ 
+} #end calc_water_year
+
+combine_flow_metrics<- function(...){
+  flow_metrics_all_sites<-bind_rows(...)
```
jds485, line 73: I think it would be better to transform this into a matrix with site numbers as rows and metrics as columns. It should reduce file size for when we have many sites to work with. If you don't want to do that with this PR, feel free to start an issue to change the combine method.

jds485 (changes requested):
This looks great overall! Some minor change requests. See suggestions/requests to move items to issues and discussions.

_targets.R
```diff
@@ -0,0 +1,83 @@
...
+              command=combine_flow_metrics(!!!.x))
+) #end list
```
jds485, line 83: It could be helpful to combine the drainage areas into a matrix of site number and drainage area. I think we'll want to divide the metrics by drainage area

_targets.R
```diff
@@ -0,0 +1,83 @@
+library(targets)
+library(tarchetypes)
+library(dataRetrieval)
+library(EflowStats)
+library(tidyverse)
+library(lubridate)
+library(fasstr)
+source("./1_fetch/src/get_nwis_data.R")
+source("./1_fetch/src/calc_HIT.R")
+
```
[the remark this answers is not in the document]
slevin75: ok , did this.

1_fetch/src/get_nwis_data.R
slevin75: Have created an issue for this.

_targets.R
```diff
@@ -0,0 +1,83 @@
...
+             clean_daily_data(p1_daily_flow,p1_screen_daily_flow,yearType="water")),
+  
+  ##get drainage area (needed for EflowStats - can change this later if DA is part of the
+  ## basin characteristics file)
+  tar_target(p1_drainage_area,
+             readNWISsite(siteNumber=sites)$drain_area_va),
+  
+  ##get peak flow data (also needed for Eflow Stats)
+  ##note - in this function, dataRetreival gives an NA for date any time the day of occurance is unknown in NWIS,
+  ##I just removed any peak that didn't have a date, but if we want, I can fill in the date so we don't lose that
```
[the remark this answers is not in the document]
slevin75: If it comes to that, I think we could use something like MOVE3 for filling in short periods of maybe a week or less, of missing data during low flow months (would not do it during higher flow months that we are interested in).  But, the more I look at different sites, it seems like when the data is missing, its really missing in substantial chunks - like 3 months are missing, so probably this won't work in those cases.

_targets.R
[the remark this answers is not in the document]
slevin75: I think in most cases the contributing drainage area will not be populated. I suspect this is mostly an issue in the northern plains where they have all the prairie potholes. But even then - the affect of the noncontributing area on runoff is very complicated and dynamic so I wouldn't really trust those numbers anyway.

_targets.R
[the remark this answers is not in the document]
slevin75: oops! ok have changed that!

1_fetch/src/get_nwis_data.R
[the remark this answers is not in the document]
slevin75: this is very odd.  I have changed the column names to discharge and discharge_cd.  I did a search for "X_" so hopefully when the names are a little wacky, they at least start with X_, otherwise this might not work.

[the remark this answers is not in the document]
jds485: I'll take a look at how you programmed the check. I think the "cd" column will always be one to the right of the "discharge" column, so it could be `grep` the colnames for "cd" and take that column index as well as the index one smaller

1_fetch/src/get_nwis_data.R
slevin75: OK, have added a couple lines in the screening function to remove any provisional data prior to checking for complete years.  If we decide to remove estimated data, I can do that here as well.

_targets.R
slevin75: I actually was having trouble with this - the drainage area targets were not combining the way I thought they should be.  And then I thought maybe if we end up getting the drainage area from the basin characteristics, there may be an easier way of doing it later .  If the basin characteristics are some sort of data frame of values or something, it would be eaiser to just grab the drainage area column.  I'm going to make this an issue and we can think about how to do it when we get the basin characteristics.

1_fetch/src/calc_HIT.R
```diff
@@ -2,7 +2,7 @@
 
 
 calc_HITmetrics<-function(data,yearType,drainArea,floodThreshold){
-  
+  suppressWarnings(calc_allHIT(x=data,yearType=yearType,drainArea=drainArea,floodThreshold=floodThreshold))
   out_data<-calc_allHIT(x=data,yearType=yearType,drainArea=drainArea,floodThreshold=floodThreshold)
```
jds485: This should be `out_data<-suppressWarnings(calc_allHIT(x=data,yearType=yearType,drainArea=drainArea,floodThreshold=floodThreshold))`

_targets.R
```diff
@@ -1,17 +1,16 @@
 library(targets)
 library(tarchetypes)
-library(dataRetrieval)
-library(EflowStats)
 library(tidyverse)
```
jds485, line 3: `library(tidyverse)` can be deleted because you're loading it below with `suppressPackageStartupMessages(library(tidyverse))`

1_fetch/src/calc_HIT.R
jds485: Should be `out_data<-suppressWarnings(calc_allHIT(x=data,yearType=yearType,drainArea=drainArea,floodThreshold=floodThreshold))`

jds485 (approved):
2 small changes to make to suppressWarnings commands. Otherwise looks good and ran for me

jds485:
You can merge after you've made those edits

1_fetch/src/calc_HIT.R
slevin75: ahh - ok, changed this.

Issue #15, opened 2021-11-24
additions to screen_daily_data function


slevin75:
This is a reminder for myself mostly to add a line in the screen_daily_data function to remove estimated as well as provisional data and also to add the site number to the output table in this function.  Should be a quick fix but I just don't want to forget over the weekend.

slevin75:
Has been fixed and merged.

jds485:
Did you fix this in your copy of the repo?

slevin75:
oh crud - yes, its in my fork but not in this version. What step did I miss? Why is this so confusing to me?!

slevin75:
ok - I forgot to merge it after you reviewed it.  I think it should be updated now.

jds485:
Thanks! I think GitHub will close these issues automatically after the merge when the issue is listed in the PR, so that's why I caught it. I just realized that we didn't add "closes #15" to that PR, so I added that to the comments

jds485:
Actually, it's in the "Linked issues" option on the right panel of the PR

PR #16, opened 2021-12-01
2 small changes to data  screening function

slevin75:
added code to remove estimated as well as provisional data and append site number to data screening output


jds485 (approved):
Looks good!

jds485:
closes #15

Issue #11, opened 2021-11-10
Combine drainage areas and site numbers into a matrix or dataframe


slevin75:
Currently drainage area targets are downloaded from NWIS and are single values, but it would be good to have them in a matrix with row names equal to the streamgage so we can select by drainage area.  We might end up getting the drainage area from the basin characteristics file, in which case it will be easier to just select the column from that file. If we continue to use NWIS values, they will need a tar_combine function to pull them all together.

Issue #9, opened 2021-11-10
Get drainage area from basin characteristics (target p1_drainage_area)


slevin75:
EflowStats requires the drainage area for some functions. Right now target p1_drainage_area is downloading the drainage area from NWIS but eventually, we will probably have a data file with all the basin characteristics and we should probably just get it from that file.

jds485:
I'll just add the comment from #2 that there is a "contributing drainage area" and a "drainage area" that are different due to connectivity information (e.g., ponding, water transfers, etc.).

So, I think we'll have at least 3 sources of drainage area (NWIS, NHD, and GAUGES) and at least 4 variables to choose from with the contributing drainage area option

ajsekell:
I would have a look and compare the 3 sources of drainage areas since it'll pretty straightforward to do. They might all be very close, but I would check. Guessing NWIS and GAGES2 will be the closest to each other.

jds485:
One way to do this comparison is to create a new target for the NHD and GAGES drainage areas to which we can compare the NWIS p1_drainage_area target

cstillwellusgs:
Don't worry about NHD drainage areas. Might make sense to use Falcone's since we are using his other features for other stuff. Compare with NWIS just to be sure.

PR #19, opened 2021-12-03
Reformat p1_drainage_area as matrix

cstillwellusgs:
#11 

Because we can potentially pull drainage areas from different sources (NWIS, Weickzorek, etc.) I reformatted the target p1_drainage_area to retain site numbers and then add drainage area as a second column (in this case, DA_NWIS). This can be further modified if we want to use the drainage areas based on COMID or another source. The calc_HITmetrics function was modified accordingly to select from the DA_NWIS column (for now).

### DIFF
--- a/1_fetch/src/calc_HIT.R
+++ b/1_fetch/src/calc_HIT.R
@@ -2,7 +2,6 @@
 
 
 calc_HITmetrics<-function(data,yearType,drainArea,floodThreshold){
-  drainArea <- drainArea$DA_NWIS
   out_data<- suppressWarnings(calc_allHIT(x=data,yearType=yearType,
                                           drainArea=drainArea,
                                           floodThreshold=floodThreshold))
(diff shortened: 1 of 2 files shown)


1_fetch/src/calc_HIT.R
```diff
@@ -2,6 +2,7 @@
 
 
 calc_HITmetrics<-function(data,yearType,drainArea,floodThreshold){
+  drainArea <- drainArea$DA_NWIS
```
jds485, line 5: Because drainage area is an argument to this function, where the function is called in targets.R we can change `drainArea=p1_drainage_area` to `drainArea=p1_drainage_area$DA_NWIS`. That way we don't have to modify this function.

jds485:
Related issue #9

jds485 (changes requested):
I don't think we need to change the function and can instead use `p1_drainage_area$DA_NWIS`

jds485 (approved):
Thanks, looks good! You can merge when ready

Issue #27, opened 2021-12-16
gage 06408700 causes calcHITmetrics to fail


slevin75:
gage 06408700 in SD causes calcHITmetrics fails and I can't figure out why. I don't think its a targets or our pipeline issue, I think its actually an Eflostats issue, so if you run the group of gages near CO, it will fail at that gage.  I haven't found other gages that have this issue yet.

jds485:
I can take a look at this issue associated with the planned work on #25

jds485:
Is this the error message that you were getting? Trying to reproduce the problem and then diagnose why it's happening
> Error: callr subprocess failed: $ operator is invalid for atomic vectors

jds485:
I found the error and it is in the EflowStats code in the `calc_timingHigh` function. In the following if statement, it's possible that all flows are flood flows. When that happens, `nonFloodEvents` = 0 and is not a data.frame as expected. I'll file this as an issue in that repo.
> if (!is.null(floodThreshold)) {
    dailyMaxFlow <- dplyr::summarize(dplyr::group_by(x, 
      day), maxFlow = max(discharge))
    nonFloodEvents <- find_eventDuration(dailyMaxFlow$maxFlow, 
      threshold = floodThreshold, aggType = "none", type = "low", 
      trim = FALSE)
    maxDuration <- max(nonFloodEvents$duration)
    th3 <- maxDuration/365
  }

jds485:
For our project, we will not need these timing metrics or several low flow metrics, so I'm proposing that we specify the `stats` argument in `calc_HITmetrics` as 
stats <- c("calc_magAverage", "calc_magLow", "calc_magHigh", "calc_frequencyHigh", "calc_durationHigh", "calc_rateChange")

jds485:
This error happening makes me very suspicious of the reported flood threshold for this gage. It's impossible to have a full year of flood flows, right?!

cstillwellusgs:
Ha! I'm not a flood expert... but that is suspicious. Another item for us to investigate is the flood threshold itself (if I remember correctly, Sara specifies some parameter = 0.6). Is this flood threshold valid across all regions, drainage area sizes, etc.?

slevin75:
@cstillwellusgs @jds485

[...]

PR #28, opened 2021-12-21
[27] Added statistics to compute for EflowStats

jds485:
This PR reduces the number of statistics computed by EflowStats. The `stats_HIT` vector in `_targets.R` specifies the functions used by EflowStats. This will solve the problem in #27, but I'm not closing that issue with this PR because of the ongoing discussion.

The most recent commit further reduces the metrics to only those we will use for the project. It also implements drainage area normalization, and removes the median and annual mean normalization used for some metrics

There are also style changes to the 3 R scripts (spaces, indentations, white space).

### DIFF
--- a/1_fetch/src/calc_FDC.R
+++ b/1_fetch/src/calc_FDC.R
@@ -16,6 +16,9 @@ calc_FDCmetrics <- function(site_num, clean_daily_flow, yearType,
   NE_flows <- quantile(data$discharge, probs = NE_probs, type = 6, names = F)
   
   #Compute metrics for each NE_flows threshold
+  #dh, fh, and vh are for duration, frequency, and volume of high flows. 
+  #fdc is for flow-duration curve. 
+  #1 and 2 are two different volume metrics
   dhfdc <- fhfdc <- vhfdc1 <- vhfdc2 <- vector('numeric', length=length(NE_flows))
   for (i in 1:length(NE_flows)){
     #Get the duration of events above each of the NE_flows (dh17 style metrics)
--- a/1_fetch/src/calc_FDC.R
+++ b/1_fetch/src/calc_FDC.R
@@ -1,0 +1,70 @@
+#Function used to compute FDC-based metrics
+calc_FDCmetrics <- function(site_num, clean_daily_flow, yearType, 
+                            drainArea_tab, NE_probs, digits = 3){
+  print(site_num)
+  data <- clean_daily_flow %>%
+    filter(site_no == site_num)
+  
+  drainArea <- drainArea_tab %>%
+    filter(site_no == site_num) %>%
+    pull(drainArea)
+  
+  #Validate using the same function as EflowStats
+  data <- validate_data(data, yearType)
+  
+  #Get the non-exceedance flow values (mh15 style metrics)
+  NE_flows <- quantile(data$discharge, probs = NE_probs, type = 6, names = F)
+  
+  #Compute metrics for each NE_flows threshold
+  dhfdc <- fhfdc <- vhfdc1 <- vhfdc2 <- vector('numeric', length=length(NE_flows))
+  for (i in 1:length(NE_flows)){
+    #Get the duration of events above each of the NE_flows (dh17 style metrics)
+    #We can use the trim=TRUE argument to not count events that start at the 
+    #beginning or end of the record (these have unknown duration)
+    dhfdc[i] <- find_eventDuration(data$discharge, threshold = NE_flows[i], aggType = "average")
+    
+    #Get the number of events above each of the NE_flows (fh5 style metrics)
+    yearlyCounts <- dplyr::do(dplyr::group_by(data, year_val), 
+                              {find_events(.$discharge, 
+                                           threshold = NE_flows[i], type = "high")
+                              })
+    yearlyCounts <- na.omit(yearlyCounts)
+    yearlyCounts <- dplyr::summarize(dplyr::group_by(yearlyCounts, year_val), 
+                                     numEvents = max(event))
+    fhfdc[i] <- mean(yearlyCounts$numEvents)
+    
+    #Get the flow volumes above the NE_flows threshold for events above each of the NE_flows (mh21 style metrics)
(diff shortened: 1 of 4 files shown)


1_fetch/src/calc_HIT.R
```diff
@@ -1,21 +1,49 @@
...
+  out_data$site_num <- unique(data$site_no)
+  
+  #Select only the metrics to save
+  if (!is.null(save_metrics)){
+    out_data <- filter(out_data, indice %in% save_metrics)
+  }
+  
+  #Normalize metrics by drainage area
+  if (!is.null(norm_DA)){
+    out_data$statistic[out_data$indice %in% norm_DA] <- out_data$statistic[out_data$indice %in% norm_DA]/drainArea
```
jds485, line 32: is there a tidy method of doing replacements of data like this?

jds485:
There's a new function for FDC-based metrics. I verified that it gives the same results as EflowStats for the metrics that are the same. These metrics are currently in a comment in _targets.R. After this is reviewed, we can add a target that combines the FDC and EflowStats metrics and drops the duplicate metrics

1_fetch/src/calc_FDC.R
```diff
@@ -0,0 +1,70 @@
...
+    yearlyCounts <- na.omit(yearlyCounts)
+    yearlyCounts <- dplyr::summarize(dplyr::group_by(yearlyCounts, year_val), 
+                                     numEvents = max(event))
+    fhfdc[i] <- mean(yearlyCounts$numEvents)
+    
+    #Get the flow volumes above the NE_flows threshold for events above each of the NE_flows (mh21 style metrics)
+    lst <- find_events(data$discharge, threshold = NE_flows[i])
+    eventData <- na.omit(lst)
+    numEvents <- length(unique(eventData$event))
+    totalFlow <- sum(eventData$flow - NE_flows[i])
```
jds485, line 40: This is how EflowStats computes the volume. I'm thinking it would be better to use the total volume: `totalFlow <- sum(eventData$flow)`

1_fetch/src/get_nwis_data.R
```diff
@@ -1,160 +1,151 @@
...
+    select(date, discharge)
   
-  filepath<-p1_peak_flow[grep(site_num,p1_peak_flow)]
-  peaks<-read_csv(filepath) 
-  peaks$site_no<-as.character(peaks$site_no)
-  df_pk<-data.frame(date=as.Date(peaks$peak_dt),peak=peaks$peak_va)
-  floodThreshold<-get_peakThreshold(df_dv,df_pk,perc=perc,yearType=yearType)
-  df_out<-data.frame(site_no=site_num,floodThreshold)
+  filepath <- p1_peak_flow[grep(site_num, p1_peak_flow)]
+  peaks <- read_csv(filepath, show_col_types = FALSE)
```
cstillwellusgs, line 131: `show_col_types = FALSE` caused an error on my end. Maybe a version issue? Not sure. 

I replaced with `col_types = cols()` and it ran successfully.

cstillwellusgs (changes requested):
Should we include another function (and target) that combines the HIT and FDC metrics into a single data frame? Or maybe we are waiting on that until we have a better idea of the next steps?

I added a few comments for you to address, mostly minor. Feel free to merge once those are addressed!

1_fetch/src/calc_FDC.R
```diff
@@ -0,0 +1,70 @@
...
+    pull(drainArea)
+  
+  #Validate using the same function as EflowStats
+  data <- validate_data(data, yearType)
+  
+  #Get the non-exceedance flow values (mh15 style metrics)
+  NE_flows <- quantile(data$discharge, probs = NE_probs, type = 6, names = F)
+  
+  #Compute metrics for each NE_flows threshold
+  dhfdc <- fhfdc <- vhfdc1 <- vhfdc2 <- vector('numeric', length=length(NE_flows))
```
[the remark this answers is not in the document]
jds485: I'll add a comment in the code. dh, fh, and vh are for duration, frequency, and volume of high flows. fdc is for flow-duration curve. 1 and 2 are two different volume metrics

1_fetch/src/calc_FDC.R
jds485: Yes, that's correct. The documentation for EflowStats says total flow, but the code is flow above threshold.

_targets.R
```diff
@@ -32,7 +32,7 @@ stats_HIT <- c("calc_magAverage", "calc_magLow", "calc_magHigh",
 metrics <- c('ma1', 'ma2', 
              'ml17', 'ml18', 
              'mh15', 'mh16', 'mh17', 'mh20', 'mh21', 'mh24', 'mh27', 
-             'fh1', 'fh2', 'fh5', 'fh8',
```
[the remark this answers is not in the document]
jds485: I checked these were identical for the test sites we're currently using. Sorry for not mentioning that in the comments

1_fetch/src/calc_FDC.R
cstillwellusgs: Interesting. Go ahead with it now, but I think we should talk this over at our next meeting.

jds485: Sounds good. I have that in our meeting agenda

1_fetch/src/get_nwis_data.R
jds485: Interesting - that error does not appear for me. I'll change to `col_types = cols()` to be safe. That is the default option and it also suppressed the printed info for me.

1_fetch/src/calc_HIT.R
jds485: Thanks! I decided to use the `case_when` option for all of these replacements in this script. I verified that it gave the same results as before.

jds485:
> Should we include another function (and target) that combines the HIT and FDC metrics into a single data frame? Or maybe we are waiting on that until we have a better idea of the next steps?

I made a new issue #29 for this. I'll need to address this before clustering, and I think I see how to do it so I'll assign myself

PR #33, opened 2022-01-13
Nonstationarity

slevin75:
Added targets which  computes the fdc metrics over a moving window for each site, plot the moving window metrics for each site and then plots summary plots of normalized metrics across all sites.

### DIFF
--- a/1_fetch/src/moving_window_functions.R
+++ b/1_fetch/src/moving_window_functions.R
@@ -33,6 +33,7 @@ plot_trend_summary<-function(moving_window_metrics,outdir){
   #for a site,the sd will be 0)
   df_norm<-moving_window_metrics %>%
     group_by(site_num,indice) %>%
+    filter(site_num %in% screened_plot_sites)%>%
     mutate(norm = (statistic - mean(statistic))/sd(statistic))%>%
     mutate(indice_grp=word(indice,start=1,sep="_") ) 
 
--- a/1_fetch/src/moving_window_functions.R
+++ b/1_fetch/src/moving_window_functions.R
@@ -12,7 +12,7 @@ calc_moving_window_metrics<-function(site_num, window_length,increment, min_yrs_
                   by = increment)
 
   ##remove any starting years which would result in a window with no data (if there are large data gaps). 
-  df_screen<-bind_rows(map(start_yrs,window_screen,data))
+  df_screen<-bind_rows(map(start_yrs,window_screen,data,window_length))
   start_yrs<-start_yrs[start_yrs %in% df_screen$start_yr[which(df_screen$years_in_window > min_yrs_in_window)]==TRUE]
   
   
@@ -34,11 +34,11 @@ plot_trend_summary<-function(moving_window_metrics,outdir){
   df_norm<-moving_window_metrics %>%
     group_by(site_num,indice) %>%
     mutate(norm = (statistic - mean(statistic))/sd(statistic))%>%
-    mutate(indice_grp=word(indice,start=1,sep="_") )%>%
-    filter(!is.na(norm))
+    mutate(indice_grp=word(indice,start=1,sep="_") ) 
+
   
   indice_grp<-unique(df_norm$indice_grp) 
-  ##map over indice_grp and produce a plot file for each group, with all the quantiles 
+ ##map over indice_grp and produce a plot file for each group, with all the quantiles 
   map_out<-map(indice_grp,make_summary_plot,
           data=df_norm,
           outdir= outdir)%>%
@@ -56,6 +56,7 @@ make_summary_plot<- function(grp,data, outdir){
     geom_smooth()+
     facet_wrap(~indice, ncol=3)
   
+
   filepath <- file.path(outdir, paste0("moving_window_summary_", grp, ".png"))
   
   ggsave(filename=filepath,
@@ -94,7 +95,7 @@ screen_plot_sites<-function(moving_window_metrics, min_windows){
   metrics_screen<-moving_window_metrics %>% 
     group_by(site_num,indice) %>%
     count()
-  "ok1"
+
   plot_sites<-metrics_screen %>%
     filter(n> min_windows) %>%
     pull(site_num)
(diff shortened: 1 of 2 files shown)


_targets.R
```diff
@@ -153,4 +164,39 @@ list(
   # dh17 = dhfdc_q0.5
   # dh20 = dhfdc_q0.75
   
+  ########moving window nonstationarity stuff
+   ##table with all the FDC metrics computed on a moving window. The parameter min_yrs_in_window
+  ##screens out any moving windows for which there are too few years to be reliable. Can be an issue
```
jds485, line 169: You can make `min_yrs_in_window` a parameter. I don't think we'll change it from 15, but it could help to have the option. And these comments describing the function parameters should be moved to the function's R script

_targets.R
```diff
@@ -46,29 +47,39 @@ metrics_med_DA <- c('mh15', 'mh16', 'mh17', 'mh21', 'mh24', 'mh27')
 #non-exceedance quantiles for additional metrics - daily flows
 NE_quants = c(seq(0.5, 0.95, 0.05), 0.98, 0.99, 0.995)
 
+
+###moving window parameters
+window_length <- 20  ##needs to be <= complete_years
+increment <- 1
```
jds485, line 53: Did you test using other increments? In targets, I think we could `map` over a vector of increments to make plots for several increments with one pipeline build

_targets.R
jds485: Same comment for `min_windows`

_targets.R
jds485: It might be `cross` instead of `map` as the pattern because I see you're mapping over sites already

1_fetch/src/moving_window_functions.R
```diff
@@ -0,0 +1,139 @@
...
+    metrics$quantile<-as.factor(word(metrics$indice,start=-1,sep="_q"))
+  
+    ##I tried a line plot here too but the loess plots were just easier to look at and
+    ##get a general idea of the trend.  The line plots got very messy looking but sometimes
+    ##the loessdoes weird things if there are lots of gaps in the data, so idk which is better.
+    p1<-ggplot(metrics,aes(start_Year,statistic,color=quantile))+
+      geom_point(size= .5) + facet_wrap(~index_grp,scales="free")+
+      geom_smooth(se=FALSE)+
+      theme(legend.position="bottom")
+    filepath <- file.path(outdir, paste0(site, "_",window_length,"yr_moving_window_plots.png"))
```
jds485, line 123: `window_length` should be a function argument

1_fetch/src/moving_window_functions.R
```diff
@@ -0,0 +1,139 @@
...
+  print(site)
+  metrics<-moving_window_metrics %>%
+    filter(site_num==site)
+  if(nrow(metrics)> 0){
+    metrics$index_grp<-word(metrics$indice,start=1,sep="_")
+    metrics$quantile<-as.factor(word(metrics$indice,start=-1,sep="_q"))
+  
+    ##I tried a line plot here too but the loess plots were just easier to look at and
+    ##get a general idea of the trend.  The line plots got very messy looking but sometimes
+    ##the loessdoes weird things if there are lots of gaps in the data, so idk which is better.
```
jds485, line 118: Let's keep as loess for now

1_fetch/src/moving_window_functions.R
```diff
@@ -0,0 +1,139 @@
...
+}
+
+screen_plot_sites<-function(moving_window_metrics, min_windows){
+
+  ##screen out any sites that do not have enough moving windows to plot
+  ##using 
+  metrics_screen<-moving_window_metrics %>% 
+    group_by(site_num,indice) %>%
+    count()
+  "ok1"
```
jds485, line 97: delete "ok1"?

1_fetch/src/moving_window_functions.R
```diff
@@ -0,0 +1,139 @@
...
+  plot_sites<-metrics_screen %>%
+    filter(n> min_windows) %>%
```
jds485, line 99: does `count()` add a column named `n`?

1_fetch/src/moving_window_functions.R
```diff
@@ -0,0 +1,139 @@
...
+  }
+}
+
+window_screen<-function(start_yr, data){
+  ##screener to count the number years of data within a window.
+  ##using this to select out any starting years that result in a window with 
+  #no data or with too few years of data to be reliable
+  
+  x<-data %>% 
+    filter((year_val >= start_yr) &(year_val < (start_yr + window_length)))
```
jds485, line 136: `window_length` should be a function argument

1_fetch/src/moving_window_functions.R
```diff
@@ -0,0 +1,139 @@
...
+               NE_probs=NE_probs,digits=3)
+  data_out<-bind_rows(map_out)
+ 
+  return(data_out)
+}
+
+
+plot_trend_summary<-function(moving_window_metrics,outdir){
+  #normalize metrics and remove NAs (when there is only 1 moving window
+  #for a site,the sd will be 0)
```
jds485, line 33: Is it possible to have NAs when `min_windows` > 1?

1_fetch/src/moving_window_functions.R
```diff
@@ -0,0 +1,139 @@
...
+
+calc_FDC_subset<- function(start_yr, window_length,site_num, clean_daily_flow, yearType, 
+                           drainArea_tab, NE_probs, digits = 3 ){
+
+ ##subset the clean_daily_flow data
+
+  sub_clean_daily_flow<- clean_daily_flow %>%
+    filter((year_val >= start_yr) &(year_val < (start_yr + window_length)))
+  
+  df_FDC<- calc_FDCmetrics(site_num,
```
jds485, line 77: Just noting that I changed the return format of this function in my open PR. I can add an option to return in this format if that's simpler to work with. The `year_start` argument is also new to allow for non-water year seasons. So there will be some merge conflicts to resolve between these 2 PRs

jds485 (changes requested):
The code and plots look good! I have some changes in comments.

_targets.R
slevin75: I did it with a 2 year increment just to make sure it worked ok but the 1 year increment didn't take an exceedingly long time, so I just left it at that. Lets talk about if this is necessary.  A plot with a longer increment will be the same as a 1-yr plot except with fewer points on it, so I'm not sure what that would tell us.

1_fetch/src/moving_window_functions.R
```diff
@@ -0,0 +1,139 @@
...
+  metrics<-moving_window_metrics %>%
+    filter(site_num==site)
+  if(nrow(metrics)> 0){
+    metrics$index_grp<-word(metrics$indice,start=1,sep="_")
+    metrics$quantile<-as.factor(word(metrics$indice,start=-1,sep="_q"))
+  
+    ##I tried a line plot here too but the loess plots were just easier to look at and
+    ##get a general idea of the trend.  The line plots got very messy looking but sometimes
+    ##the loessdoes weird things if there are lots of gaps in the data, so idk which is better.
+    p1<-ggplot(metrics,aes(start_Year,statistic,color=quantile))+
```
jds485, line 119: Can you add the gage number to these plots? Maybe as a title

_targets.R
jds485: Okay, thanks, that makes sense. We can leave it as 1-yr then

1_fetch/src/moving_window_functions.R
slevin75: yes it does

1_fetch/src/moving_window_functions.R
slevin75: Hmm, probably not unless all the metrics were identical.  I took it out.  I probably had this in there before I decided to make min_windows a thing.

jds485 (approved):
Looks good! It runs for me with the new edits. Did you receive this warning for p1_moving_window_summary_plots? And do you know what causes this problem? I didn't look into it yet.
"Removed 119 rows containing missing values geom_point"

slevin75:
I just looked into the warning about the removed points.  It is because when I made the summary plots, I didn't use the minimum window criteria, so there are some sites in there that had one or 2 moving window and so it couldn't normalize because the standard deviation was 0.  I added a filter line to only use those same sites that are in the single-site plots, so you shouldn't get that error now.

jds485:
Good catch!

PR #46, opened 2022-01-29
Logs to track changes to dataRetrieval requests

cstillwellusgs:
I used Lauren's example code from the drb-inland-salinity repo. I might still be missing something... as it is now, the log files will overwrite previous logs. Lauren mentioned something about using diff to compare old and new logs, but I'm not sure if we want to keep a running log (do not overwrite old logs) for ease of access down the road?


jds485:
This looks good! I think we should track all data pulls from NWIS. Looks like drainage area and flood threshold for this repo

1_fetch/src/get_nwis_data.R
```diff
@@ -34,6 +34,26 @@ get_nwis_daily_data <- function(site_num, outdir, parameterCd, startDate, endDat
...
+  daily_flow_list <- map(files_in, read_csv, 
+              col_types = cols(agency_cd = col_skip(), 
+                               site_no = col_character(), 
+                               Date = col_date(format = "%Y-%m-%d"), 
+                               discharge = col_double(), 
+                               discharge_cd = col_character()))
+  daily_flow_df <- bind_rows(daily_flow_list)
+  daily_flow_log <- daily_flow_df %>%
+    group_by(site_no) %>%
+    summarize(num_days = n(),
```
jds485, line 48: Could you add a column for number of provisional or estimated days? Maybe `num_Pe` for the name

1_fetch/src/get_nwis_data.R
```diff
@@ -34,6 +34,26 @@ get_nwis_daily_data <- function(site_num, outdir, parameterCd, startDate, endDat
...
+                               Date = col_date(format = "%Y-%m-%d"), 
+                               discharge = col_double(), 
+                               discharge_cd = col_character()))
+  daily_flow_df <- bind_rows(daily_flow_list)
+  daily_flow_log <- daily_flow_df %>%
+    group_by(site_no) %>%
+    summarize(num_days = n(), 
+              start_date = min(Date), 
+              end_date = max(Date), 
+              mean_flow = mean(discharge, na.rm = TRUE),
```
jds485, line 51: Similar comment: can you add mean and sd for flows that are not P or e?

_targets.R
```diff
@@ -63,7 +63,8 @@ min_windows <- 10  ##Must have this many windows available in order to plot
```
jds485, line 63: at the top of the `_targets.R` script, add a `dir.create` for `1_fetch/out/logs/`

jds485:
> using diff to compare old and new logs

Yes, we'll commit the log files to GitHub. You can commit the log files you generated in this PR. The purpose of the log files is mostly to detect when NWIS data have changed.

> but I'm not sure if we want to keep a running log (do not overwrite old logs) for ease of access down the road?

I think we'd need other log files for tracking the model and input data versions (related to #1). The log files in this PR are tracking raw data changes.

jds485 (changes requested):
I didn't run this yet, but it looks good! See minor change requests in previous comments. I'll run it after you've addressed those

cstillwellusgs:
I think I updated the .gitignore file correctly to track the log files....

Addressed all your comments except creating a log for flood threshold, which I think is a function from the EFlowStats package (not dataRetrieval).

.gitignore
```diff
@@ -1,8 +1,10 @@
 #Targets
 _targets
 
-#All output
-*/out/*
+#All output except dataRetrieval logs
+*/out/stationarity_plots
+*/out/*_dv.csv
+*/out/*_pk.csv
```
jds485, line 7: You could also use an exception to the rule. Like:
```
*/out/*
!*/out/logs
```

jds485: I addressed this in a commit

jds485 (approved):
Looks good and runs for me. Just one change requested to add the mean and sd for approved data to the log file. You can merge after that. Also note that I added a commit, so you'll need to do a git pull

PR #48, opened 2022-02-04
Check GagesII sites w/o peak flow

cstillwellusgs:
REMINDER - DO NOT MERGE

Add code (and log with summary) to identify extra sites with daily flow data but without peak flow data.


_targets.R
```diff
@@ -81,202 +81,208 @@ gagesii$ID <- substr(gagesii$ID, start=2, stop=nchar(gagesii$ID))
...
-                                window_length = window_length,
-                                outdir = "1_fetch/out/stationarity_plots"),
-             map(p1_screened_plot_sites),
-             format = "file"),
-  
-  tar_target(p1_moving_window_summary_plots,
-             plot_trend_summary(moving_window_metrics = p1_moving_window_metrics,
-                                screened_plot_sites = p1_screened_plot_sites,
-                                outdir = "1_fetch/out/stationarity_plots"),
+  tar_target(p1_has_data_check,
```
jds485, line 97: Instead of needing to comment out targets, you can run `tar_make(<target name>)` to run only the target you want and its dependencies

jds485:
Thanks for making this summary. We can discuss on Monday, but I think it's okay to drop these sites instead of programming functions to retain them for the FDC metric calculations. I'll leave this open until after our meeting

cstillwellusgs:
Not merging - discussed on 02/07/2022. Only 8 sites without peaks that contain daily flows (out of > 2000); determined that this was a 'non-issue'. Closing without merging.

Issue #25, opened 2021-12-13
Analysis of metric-based study regions


jds485:
Task list. These may be several PRs (e.g., 1st, then 2nd-end)
- [x] Create function that computes FDC-based metrics for flows from median and larger. Addressed in #28 
- [x] Create function for seasonal analysis of FDC-based metrics. Addressed in #30 
- [x] Create function for cluster analysis of resulting seasonal distributions. Addressed in #50 
- [x] Create function that visualizes resulting cluster regions. Addressed in #50 
- [ ] Create function that generates the union of selected regions and saves to shapefile. Likely will move to a new issue to explore options for combining point-based clusters into watershed-based regions

jds485:
For seasons in which we have no events, do we want the duration of events above a % threshold for that season to be reported as 0 or as NA? The resulting seasonal average (computed across all years for a site) will be affected by 0s, but will not be affected by NAs (NAs are dropped). Treating as NA will provide a more accurate estimate of the average duration of events above the threshold in that season, so I'm leaning towards using NAs

jds485:
Just realizing that the choice does not matter if we transform the raw seasonal durations into fractions of a year: `(season duration)/(annual total duration)`

jds485:
I can see 2 approaches for computing the resulting seasonal metrics and I'm not sure which is best to use:
1. Period of record method: 
  - Compute the seasonal durations for each year (4*n years values)
  - Average the seasonal values over all years (4 values)
  - Compute the fraction per season (4 values)
      
2. Annual timeseries method:
  - Compute the seasonal durations for each year (4*n years values)
  - Compute the fraction per season (4*n years values)
  - Average the fractions over all years (4 values)

Any suggestions? @FlowChar, what approach did you use for previous work with seasonal metrics?

cstillwellusgs:
Not sure I fully follow - what do you mean by "compute the fraction per season" (in both approach 1 and 2)?

jds485:
For example, considering number of events:
fraction per season = events in season / total events for the year

jds485:
I'm opting to close this. We are currently using separate regions for flows 50-70th percentile and flows 75th-95th percentile

PR #50, opened 2022-02-15
25 cluster regions

### LINKED ISSUE #40: Data are loaded in two functions. We should make a separate data processing function
The fetched flow data are loaded in `screen_daily_data` and `clean_daily_data`. We should move the data processing within `screen_daily_data` to a process function, and pass that processed data to `screen_daily_data` and `clean_daily_data`.

### LINKED ISSUE #41: Implement parallelization across sites for targets that process data
We should parallelize the processing of targets across sites. I vaguely remember that data downloads should not be parallelized, so we'll have to check how to implement the built-in `targets` parallelization options for a specified list of targets

### LINKED ISSUE #45: Investigate why ml17 is NaN for some sites
EflowStats metrics ml17 (Baseflow index) and ml18 (CV of ml17) are NaN or NA for 125 sites. Here are example sites:
image

Should look into the `calc_magLow` EflowStats function to diagnose why, and determine if a fix should be made to not report NaN or NA.

### LINKED ISSUE #51: Drop 5 GAGES 2.1 sites
From [discussion here](https://github.com/USGS-R/regional-hydrologic-forcings-ml/pull/49#issuecomment-1031877398), these gages are in locations that would likely not be representative of other catchments and may not have NHD attributes, so it's recommended to drop them.

Gages are: 02084557, 09406300, 09512200, 10143500, 10172200

jds485:
This PR addresses two tasks of #25: Cluster analysis for the FDC-based metrics and visualization of cluster analysis diagnostics and resulting clusters. You may need to install R packages listed in our `_targets.R` script.

Because this is applied to the full GAGES 2.1 dataset, it is helpful to run the pipeline in parallel with `tar_make_clustermq(workers = <number of cores>)` to speed up the data processing. The `deployment` argument in `tar_target` states whether or not to deploy on the main core or on a worker core. Addresses part of #41.

There are a few fixes to EflowStats metrics in this PR. You'll need to pull the latest version of EflowStats using the instructions in the README. Closes #45.

Added a prescreening of data function to download data only once. Closes #40

Dropped 5 gages based on discussion in #51. Closes #51 

Ideas for improving the EDA of the cluster analysis. Feel free to add and let me know which would be most helpful:

- [ ] Make a panel plot of the seasonal barplots by cluster. Maybe save with the map side-by-side and make the colors of the bars match the map point colors. Moved to #56 
- [ ] Order the plotted clusters in increasing order instead of alphanumeric. Moved to #56 
- [ ] Add the number of sites in a cluster to the barplots and nonstationary scatterplots. Moved to #56 
- [ ] Add an argument to the cluster analysis function that allows for additional features to be added to the observation matrix. Moved to #56 
- [ ] Change the silhouette diagnostic metric to a histogram of the suggested optimal number of clusters from 30 NbClust metrics. Moved to #55 
- [ ] Create cluster analysis targets for high flow seasons (high flow seasonal FDC metric targets already exist)


slevin75:
How long did this take to run on your computer?  I think it is sort of breaking mine.  I've been coaxing it along for 5 days now.  It runs for a few hours and then it just hangs and I have to stop it, shut down R studio and restart it again.  I don't know if its a memory issue what.  It is currently at the p1_moving_windows target and has finished most of the cluster functions, so I may just stop it there and look at those to get this review done, but we might need to think about how to progress  because when this gets pushed to the main branch, it will be difficult for me to add any more code if I can't really run the pipeline.

jds485:
> How long did this take to run on your computer?

Wow, that's a long time! It runs in a few hours on mine when building from scratch, and most of that time is spent downloading. It could be a memory issue because this is using all of the gages. Did you check your computer's Task Manager while it was running to see how much RAM was being used? 

Are you running the pipeline with `tar_make` or `tar_make_clustermq`? You could also try running with the specific target name that you want to build so that only that target and dependency targets build. Something like `tar_make(all_of("p3_cluster_map_png"))`. `all_of` is used to select all branches in dynamic branch targets.

In any case, sounds like it's time for us to start using the HPC. I'll file an issue for that in our repo.

3_cluster/src/seasonal_metric_cluster.R
```diff
@@ -0,0 +1,275 @@
...
+    a <- df[df$metric == unique(df$metric)[i],]
+    df_max[i,] <- a[a$ac == max(a$ac),]
+  }
+  
+  return(df_max)
+}
+
+#Function to make cluster diagnostic panel plot
+plot_cluster_diagnostics <- function(clusts, metric_mat, 
+                                     dist_method = 'euclidean',
```
slevin75, line 60: How are we going to choose the optimal number of clusters? Are they going to be based on these diagnostic plots in some automated way?  Looking at some of the plots, the place where it draws a vertical line for gap statistic seems a little off sometimes. I guess it just chooses the maximum value, but sometimes when the values plateau and then gradually increase, it just puts that line way out at 20 or something when probably it should be closer to the inflection point.  Anyway - just something to think about -  if we are going to be choosing our final number of clusters from these then we probably should make sure we agree with how those vertical lines are drawn.

slevin75 (approved):
Ok - sorry this took so long.  There were a lot of things from previous PRs that I hadn't seen yet so I just wanted to step through it to understand it.  It never did completely finish running on my my machine but it got through most of the cluster targets. Everything that did run looked ok.  It didn't get to the maps or some of the moving window plots but in the interest of moving on, I am just going to assume they run ok.

3_cluster/src/seasonal_metric_cluster.R
```diff
@@ -0,0 +1,275 @@
...
+    dev.off()
+  }
+  
+  return(fileout)
+}
+
+
+#Function to plot the cuts in trees from kmin to kmax clusters
+#clusts is the output from hclust
+plot_cuttree <- function(clusts, kmin, kmax, seq_by, dir_out){
```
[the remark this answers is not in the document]
jds485: No, I was debating including it in the diagnostic plot but decided against it. It plots rectangles on the dendrogram around each cluster of gages for a total number of clusters equal to k. I'll leave it included because it can be helpful to visualize the cluster groups after we select how many clusters to use

3_cluster/src/seasonal_metric_cluster.R
jds485: The vertical lines in the gap statistic plots are currently set to take the global maximum, which must be based on the bootstrapped mean value and not consider the uncertainty. All of the built-in methods for drawing those vertical optimality lines do not place them where I would place them considering the uncertainty, as you point out. I think the range of optimality is from maybe 8-15 clusters for most metrics. There are many other statistics that can be evaluated to assess the optimal number of clusters (30 in NbClust) and I think a histogram of the optimal number from those 30 metrics would be helpful to justify our final choice. I can add making that histogram as an issue if that sounds good.

PR #59, opened 2022-02-28
55 cluster diagnostics

### LINKED ISSUE #55: Add cluster analysis diagnostics
In the diagnostic plots, change the silhouette diagnostic metric to a histogram of the suggested optimal number of clusters from 30 NbClust metrics. Save the metrics as a dataframe of rows = flow metrics, columns = cluster diagnostics

jds485:
This PR adds a function to compute 26 NbClust cluster diagnostic metrics. I also moved the computation of the gap statistic (27th metric) out of the plotting function and into this function. The function saves the diagnostics to a target that we can access them for reference later. 

Note that it takes a long time to compute these diagnostics for all metrics, so it's probably best to evaluate this PR for just one of them using something like `tar_make(p3_FDC_cluster_diagnostics_e309a6a3)`. For reference if you want to try building in parallel with `tar_make_clustermq`, it used about 0.5 GB of RAM per worker, and each set of diagnostics took about 5 minutes to compute.

This PR also updates the diagnostic panel plot by replacing the silhouette distance metric plot with a histogram of the optimal number of clusters suggested from these 26 metrics.

Closes #55

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@ _targets
 #Clustermq worker log files
 *.log
 
+#R plots
+Rplots.pdf
+
 # History files
 .Rhistory
 .Rapp.history
--- a/3_cluster/src/seasonal_metric_cluster.R
+++ b/3_cluster/src/seasonal_metric_cluster.R
@@ -55,8 +55,39 @@ select_cluster_method <- function(clusts){
   return(df_max)
 }
 
+#Function to compute cluster diagnostics
+#kmin, kmax - min and max number of clusters to use
+#alpha - significance level
+#boot - number of bootstrap replicates
+#index - the NbClust index to compute. 'all' computes all except those with long compute times.
+compute_cluster_diagnostics <- function(clusts, metric_mat,
+                                        kmin, kmax, alpha, boot = 50,
+                                        index = 'all',
+                                        dist_method = 'euclidean',
+                                        clust_method = 'ward.D2'){
+  clusts <- clusts[[clust_method]]
+  
+  #Select all of the seasonal columns for this metric
+  metric_mat <- metric_mat[, grep(x = colnames(metric_mat), pattern = paste0(clusts$metric,'_'))]
+  
+  #Compute NbClust cluster diagnostics
+  nbclust_metrics <- NbClust::NbClust(data = metric_mat, diss = NULL, 
+                                      distance = dist_method, 
+                                      min.nc = kmin, max.nc = kmax, 
+                                      method = clust_method, index = index, 
+                                      alphaBeale = alpha)
+  
+  #Compute gap statistic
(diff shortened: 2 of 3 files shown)


jds485:
@slevin75 : try running this with `tar_make(p3_FDC_clusters_e6724caf)` to compute the clusters for only "vhfdc2_q0.995", then `tar_make(p3_FDC_cluster_diagnostics_f94089e0)` to compute the diagnostics for one target, and then `tar_make(p3_FDC_cluster_diagnostics_png_87171bfd)` to get the plot for that target. I'm not sure if this will work - it assumes the hashes are the same in our builds. Let me know if you have trouble building this.

slevin75:
> d about 0.5 GB of RAM per worker, and each set of diagnostics took about 5 minutes to compute.
> 
> This PR also updates the diagnostic panel plot by replacing the silhouette distance metric plot wi

no - it didn't work.  It didn't throw an error, it just ended the pipeline without doing anything.

slevin75:
> > d about 0.5 GB of RAM per worker, and each set of diagnostics took about 5 minutes to compute.
> > This PR also updates the diagnostic panel plot by replacing the silhouette distance metric plot wi
> 
> no - it didn't work. It didn't throw an error, it just ended the pipeline without doing anything.

I might try changing the NE_quants parameter to just 0.995 or something, so it is only doing that quantile. I don't know how much that will help.  Or I could just subset the gagesii and only use half of the CONUS or something like that. That for sure would cut the time I think.

slevin75 (approved):
This looks good.   I was able to reproduce the diagnostics plot in your comment. I ran tar_make for the p3_FDC_cluster_diagnostics_png and the  p3_cluster_map_png since I didn't get to see them last time.  Using only one quantile for NE_quants, it took about 5 hours to run, which is considerably better than 6 days!

jds485:
Thanks for the review! Great suggestion to review for only one quantile. I'm glad that worked out.

> it took about 5 hours to run

Wow! Definitely feeling the need to prioritize migrating to the HPCs. Was that 5 hrs to build the whole pipeline, or only the cluster analysis targets + dependencies?

slevin75:
> Thanks for the review! Great suggestion to review for only one quantile. I'm glad that worked out.
> 
> > it took about 5 hours to run
> 
> Wow! Definitely feeling the need to prioritize migrating to the HPCs. Was that 5 hrs to build the whole pipeline, or only the cluster analysis targets + dependencies?

That was just the cluster analysis targets and dependencies.  Downloading the streamflow took about 90 minutes, then the screening of seasonal and daily flows each took about 45 minutes, the clean daily flows and seasonal flows and the FDC metrics and FDC metrics seasonal  took about 20-30 minutes for each of those targets.

PR #66, opened 2022-03-04
65 add attributes

### LINKED ISSUE #65: Add attribute data

ajsekell:
adding sb data download and merge to gages functions.

1 make table of sb items to download
2 download and merge to list of gages

closes #65


_targets.R
```diff
@@ -253,6 +259,19 @@ list(
...
   ),
   
+  ##generate table of data to download from sciencebase
+  tar_target(p1_make_sb_dl_table,
+             makedltable(idtostart, outdir = "./1_fetch/out"),
+             format = "file"),
+  
+  ##generate table of landscape data for gages list  
+  tar_target(p1_make_sb_landscapedata,
+             downloadchildren(gages=gagesii, dldir = "./1_fetch/out/dldir",
```
cstillwellusgs, line 269: There isn't actually a dataframe called 'gagesii' just the 'gagesii_path'. In there very first target (lines 106-110), there is an internal function that reads the filepath and does some quick manipulation. I would suggest creating a new data frame (named gagesii) using that same function from lines 106-110, placing it after line 92, then replacing the internal function in the first target with the data frame callout. Then we can use the callout here as well.

_targets.R
```diff
@@ -253,6 +259,19 @@ list(
...
+             makedltable(idtostart, outdir = "./1_fetch/out"),
+             format = "file"),
+  
+  ##generate table of landscape data for gages list  
+  tar_target(p1_make_sb_landscapedata,
+             downloadchildren(gages=gagesii, dldir = "./1_fetch/out/dldir", 
+                              workdir = "./1_fetch/out/workdir",
+                              outdir = "./1_fetch/out",
+                              table_sb_dl = p1_make_sb_dl_table),
+             format = "file"),
```
cstillwellusgs, line 273: @jds485 The target 'p1_make_sb_dl_table' is pretty quick, so I'm guessing that would be deployment = 'main'. The target 'p1_make_sb_landscapedata' is slower, so I'm tempted to assign deployment = 'worker' but I suspect it will not be that easy. Would like to discuss this with you.

1_fetch/src/get_sb_data.R
```diff
@@ -0,0 +1,170 @@
+getbabies <- function(idin)
```
cstillwellusgs, line 1: Simple formatting edit - we are using snake_case (I think that is what it is called) so lets change this from 'getbabies' to 'get_babies'. Same for other function and variable names throughout.

1_fetch/src/get_sb_data.R
```diff
@@ -0,0 +1,170 @@
...
+#renaming some columns and getting rid of that ugly first column
+colnames(tabledl) <- c('id', 'parent_title', 'title', 'lastUpdated')
+rownames(tabledl) <- NULL
+
+#if you would like to export the table as csv
+filepath <- file.path(outdir, paste0("_sb_dltable.csv"))
+write.csv(tabledl,filepath, row.names = FALSE)
+return(filepath)
+}
+
```
cstillwellusgs, line 94: With a quick scroll is is hard to tell when the 'makedltable' function ends and the 'downloadchildren' function begins. I wonder if you just select all, copy, paste, if the spacing/tabs will align for a easier to read output?

1_fetch/src/get_sb_data.R
```diff
@@ -0,0 +1,170 @@
...
+downloadchildren <-function(gages, dldir, workdir, outdir, table_sb_dl)
+{
+  #making a copy to work with
+  data_at_gages <- gages
+  
+  itemfails <- "1st"
+  
+  for (row in 1:nrow(table_sb_dl))
+  {
+    item <- table_sb_dl[row, 'id']
```
cstillwellusgs, line 104: In your stand-alone version, I think this command (item <- table_sb_dl[row, 'id']) returned a character string. Since we saved the table_sb_dl as a .csv and then referenced it from targets, it reads in as a data frame (thus causing a read error). I solved this by putting everything to the right of '<-' inside 'as.character()' and it ran successfully, but I want you to make sure this is the correct way to address the issue.

1_fetch/src/get_sb_data.R
```diff
@@ -0,0 +1,170 @@
...
+  {
+    failist <- as.data.frame(itemfails[-1])
+    colnames(failist) <- c('filename')
+    rownames(failist) <- NULL
+  }
+  filepath1 <- file.path(outdir, paste0("_sb_landscapedata.csv"))
+  write.csv(data_at_gages,filepath1, row.names = FALSE)
+  filepath2 <- file.path(outdir, paste0("_sb_itemfails.csv"))
+  write.csv(itemfails,filepath2, row.names = FALSE)
+  return(filepath1, filepath2)
```
cstillwellusgs, line 169: Turns out we cannot have multiple returns. I would just pick one, I don't think I have a preference.

1_fetch/src/get_sb_data.R
```diff
@@ -0,0 +1,170 @@
...
+  
+  if (length(itemfails) > 0)
+  {
+    failist <- as.data.frame(itemfails[-1])
+    colnames(failist) <- c('filename')
+    rownames(failist) <- NULL
+  }
+  filepath1 <- file.path(outdir, paste0("_sb_landscapedata.csv"))
+  write.csv(data_at_gages,filepath1, row.names = FALSE)
+  filepath2 <- file.path(outdir, paste0("_sb_itemfails.csv"))
```
cstillwellusgs, line 167: Since we just created a single file, the file names are simply '_sb_landscapedata.csv' and '_sb_itemfails.csv'. I would maybe think of a simple naming convention for everything related to these targets (including function names, variable names, file names, etc.) and make sure you pick something consistent and easy to understand for everything. Keep in mind that we might call these functions later for the ungaged basins too, so it might help to indicate that these data are for the gages_ii watersheds (maybe we include 'outdir' and 'filename' both as variables up top so we can name this one 'blah_blah_gages_ii.csv' and a future one 'blah_blah_ungaged_comids.csv' or something like that.

cstillwellusgs (changes requested):
@ajsekell I've provided some comments to your different scripts, let me know if you have any questions. We can also delete the /working folder and all of its contents. We can walk through the comments and go through the edits together via screen share if you'd like.

1_fetch/src/get_sb_data.R
jds485: you can return multiple items if you need to. `return(list(file1 = filepath1, file2 = filepath2))`

cstillwellusgs:
@ajsekell I had to make a few minor changes. It is now up to date. You can press the Merge Pull Request button when you are ready!

Issue #56, opened 2022-02-25
Cluster analysis function enhancements


jds485:
- [ ] Make a panel plot of the seasonal barplots by cluster. Save side-by-side with the map of clusters across CONUS and make the colors of the bars match the map point colors.
 - [ ] In panel plots, order the plotted clusters in increasing order instead of alphanumeric
 - [ ] Add the number of sites in a cluster to the barplots and nonstationary scatterplots
 - [ ] Add an argument to the cluster analysis function that allows for additional features to be added to the observation matrix

PR #69, opened 2022-03-08
Aggregated cluster analysis

### LINKED ISSUE #68: Create aggregated clusters by flow quantile
Applying cluster analysis to all quantiles for all flow metrics results in many possible cluster regions. It looks like aggregating by quantile (13 total analyses, one for each flow quantile), and then aggregating into 2 sets of quantiles would be possible.

jds485:
This PR adds targets for cluster analysis by quantile (13 analyses for the 13 quantiles in NE_quants), and by two sets of aggregated quantiles (0.5 - 0.7, 0.75-0.95).

The main edits are to the cluster analysis functions. The `quantile_agg` argument is added to most of these functions to enable a simple search for all quantiles provided in the vector `metric` and create a table with all streamflow metrics computed for those quantiles. That table is used for cluster analysis. For example, it has 80 columns for aggregated quantiles 0.5 - 0.7: 4 seasons * 4 streamflow metrics * 5 quantiles.

It was a little complicated to generalize the `plot_seasonal_barplot` function for use with these aggregated quantiles. I made several helper functions to make it easier to read that function. Let me know if you think more are needed.

The nonstationarity function is not yet set up to run with the aggregated clusters. It will need similar edits made to the `plot_seasonal_barplot` function, so I think it's best to wait to set up the nonstationarity function until you review the barplot function.

To test the functions, I think it will be fastest to run the pipeline with `tar_make(p3_seasonal_barplot_clusters_quants_agg_png)` and `tar_make(p3_cluster_map_quants_agg_png)`. That will only do 2 cluster analyses and make plots for k = 3, 7, 11, and 15 clusters. Example figures:

0 75,0 8,0 85,0 9,0 95_k7_map

0 75-0 8-0 85-0 9-0 95_ward D2_diagnostics

SeasonalBarplot_0 75,0 8,0 85,0 9,0 95_k7_Metric_dhfdc

Addresses some tasks in #56 to enhance the resulting figures.

Closes #68


slevin75:
I am getting an error at the p1_screen_daily_flow_season target for gage 04153300.  The problem seems to be that this gage only has data for 2021 and 2022 which is after the end date, so when the screening function opens the file, there is nothing there.  Its not screened out in the has_data target because that target only looks to make sure data exists, but not the start or end dates.  For now, I am just adding this gage to the drop_gages list to keep the PR going.  I think it will be an easy fix to add a filter to the dv_screen variable in the has_data_check function to screen out any sites that start later than our end date.  This may pop up again if new sites gages go online.

jds485:
> The problem seems to be that this gage only has data for 2021 and 2022 which is after the end date, so when the screening function opens the file, there is nothing there.

I think @cstillwellusgs addressed this general problem in #73 by adding a check for no data, but I'm not sure if it's also checking for dates that we specify. I'll reference your comment in that PR so we can check that gage is handled correctly before it's merged. For now, dropping to run this PR sounds good!

slevin75 (approved):
This all looks good. I was able to run it and reproduce your maps and figures.  In order to get past that one site with data starting in 2021, I added the filter below to the has_data_check function.  It should remove any other sites that come online in the future as well and only have very recent data.  It can probably just be combined with the filter that is already in the function.  Anyway, it is a quick fix and I wasn't sure if you want to just add it now or save it for a different PR. 

  dv_screen <- dv_screen %>% 
    filter(begin_date < endDate)

jds485:
> In order to get past that one site with data starting in 2021, I added the filter below to the has_data_check function. It should remove any other sites that come online in the future as well and only have very recent data. It can probably just be combined with the filter that is already in the function. Anyway, it is a quick fix and I wasn't sure if you want to just add it now or save it for a different PR.

Thanks! Charlie added a PR that handles this and uses nearly the same filter

PR #70, opened 2022-03-11
Tidyverse update

cstillwellusgs:
Changes 'col_select' (deprecated in latest tidyverse version) to 'select()' within the prescreen_daily_data function.


jds485 (approved):
Looks good and it also works in tidyverse version 1.3.1. You can merge when ready.

PR #77, opened 2022-03-22
update to use p1_sites_g2 as target

### LINKED ISSUE #76: Missing or different targets in main since Mar 2
In reviewing #75, I noticed that some targets that were in USGS-R main are now missing or different. Example: the p1_sites_g2 target is different.

Looking at the [_targets.R history](https://github.com/USGS-R/regional-hydrologic-forcings-ml/commits/main/_targets.R), the change happened on Mar. 2 when Andrew's branch was merged to main. I'm not sure of the best way to handle this. Ideally, we could merge from the Mar 2 commit into the current commit on main. I can look into this tomorrow morning.

jds485:
Moves reading of GAGES II file to the target, and uses the target name in future function calls.

Closes #76

### DIFF
--- a/_targets.R
+++ b/_targets.R
@@ -91,12 +91,6 @@ gagesii_path <- "Gages2.1_RefSiteList.xlsx"
 #pipeline, ditch, etc.
 drop_gages <- c('02084557', '09406300', '09512200', '10143500', '10172200')
 
-#read in gagesII excel file for use in functions
-gagesII <- read_xlsx(gagesii_path) %>% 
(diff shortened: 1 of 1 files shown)


cstillwellusgs:
Ah. This makes sense. I asked Andrew to change it since we were going to be calling the gages_ii.xlsx multiple times, but my approach didn't track it within the targets framework. Still getting used to this...

This worked, will cause some merge conflicts with my other outstanding PRs, but I don't think it will be an issue. 

Ready to merge.

PR #81, opened 2022-04-08
Monthly avg weather

cstillwellusgs:
Since target 'p1_sb_data_g2_csv' has two filepaths (one for the actual data frame, another listing fails from sb download and merge) I had to specify the input for this target as 'p1_sb_data_g2_csv[[1]]'. It works, but I'm not sure if it is a _best practice_. If you don't like this, I could edit target 'p1_sb_data_g2_csv' to only return a single data frame, then the list of fails would be a separate target(?). Let me know what you think.

### DIFF
--- a/1_fetch/src/get_sb_data.R
+++ b/1_fetch/src/get_sb_data.R
@@ -218,3 +218,42 @@ download_children <-function(sites, sb_table_reduced, dldir, workdir, outdir, ou
   return_df = c(filepath1, filepath2)
   return(return_df)
 }
+
+calc_avg_monthly_weather <- function(sb_data, arg2) {
+  sb_data <- read_csv(sb_data[[1]], show_col_types = FALSE)
+  precip <- sb_data %>%
+    select(COMID, contains("_PPT_")) %>%
+    pivot_longer(!COMID, names_to = "unit_month_year", values_to = "precip") %>%
+    mutate(unit = str_sub(unit_month_year, 1, 3), 
+           month_name = str_sub(unit_month_year, 9, 11)) %>%
+    group_by(COMID, unit, month_name) %>%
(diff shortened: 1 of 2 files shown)


1_fetch/src/get_sb_data.R
```diff
@@ -217,4 +217,45 @@ download_children <-function(sites, sb_table_reduced, dldir, workdir, outdir, ou
...
+                                 month_name == "MAR" ~ 3, month_name == "APR" ~ 4, 
+                                 month_name == "MAY" ~ 5, month_name == "JUN" ~ 6, 
+                                 month_name == "JUL" ~ 7, month_name == "AUG" ~ 8, 
+                                 month_name == "SEP" ~ 9, month_name == "OCT" ~ 10, 
+                                 month_name == "NOV" ~ 11, month_name == "DEC" ~ 12),
+           name = paste0(unit, "_PPT_", month_name)) %>%
+    arrange(COMID, unit, month_num) %>%
+    select(COMID, name, avg_monthly_ppt) %>%
+    pivot_wider(names_from = "name", values_from = "avg_monthly_ppt")
+  temperature <- sb_data %>%
```
jds485, line 242: It looks like a lot is repeated for temperature. Could make a function with arguments for name ("_TAV_"), values_to = "temp", column to use for mean, and the new column name (avg_monthly_temp)

jds485 (changes requested):
Looks good! I have some minor suggestions in comments.

Could you paste the head(p1_monthly_weather_g2) as a comment so I can see what it looks like?

jds485 (approved):
I did some quick checks on the table of attributes. We can visualize the timeseries in a separate issue. I think it would help to make plots by cluster region. Example: to compare the avg. monthly precip timeseries to the metric seasonality in each cluster

PR #91, opened 2022-04-15
added imports of packages 

### LINKED ISSUE #72: Track package function changes as targets dependencies
For packages that we're importing from GitHub, let's track changes to the package functions in targets using:

[](https://books.ropensci.org/targets/packages.html#cb82-4)[](https://books.ropensci.org/targets/packages.html#cb82-5)[](https://books.ropensci.org/targets/packages.html#cb82-6)tar_option_set(
  packages = c("package1", "package2")
  imports = c("package1", "package2")
)

If we update the package, the pipeline will rebuild targets that depend on those functions.

@cstillwellusgs: I think if we had this in place for `tidyverse`, targets in #67 would have been labeled as outdated immediately after you updated tidyverse. So, maybe it's a good idea to include `imports` for all of the major packages that we're using.

jds485:
I added the packages to the targets `imports` argument to track changes in the functions in these packages. There were several packages that resulted in an error when building the pipeline, and therefore could not be tracked. 

For the other packages, I verified that these functions are tracked by looking at the tar_visnetwork. It has a very long tail of functions from the packages (the ones we do not use in our pipeline).
image

I think `tar_glimpse` will now be a more useful way to visualize the pipeline connections:
image

Closes #72

### DIFF
--- a/_targets.R
+++ b/_targets.R
@@ -10,7 +10,11 @@ library(tidyverse)
 tar_option_set(packages = c("fasstr", "EflowStats", "dataRetrieval",
                             "lubridate", "cluster", "factoextra", "NbClust",
                             "sf", "cowplot", "gridGraphics", "stringi",
-                            "dendextend", "scico", "tidyverse", "nhdplusTools","sbtools"))
+                            "dendextend", "scico", "tidyverse", "nhdplusTools",
+                            "sbtools"),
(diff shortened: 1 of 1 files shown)


slevin75 (approved):
This looks fine to merge.  I didn't run the entire pipeline, just a few targets.

Issue #29, opened 2021-12-28
Combine HIT and FDC metrics into a single data frame


jds485:
Include another function (and target) that combines the HIT and FDC metrics into a single data frame. 

_Originally posted by @cstillwellusgs in https://github.com/USGS-R/regional-hydrologic-forcings-ml/pull/28#pullrequestreview-840524789_

jds485:
When we used static branching, a function combined the HIT metrics for sites into a data frame with rows as sites and columns as metrics. For dynamic branching, that function will also work (see an example in my comment [here](https://github.com/USGS-R/regional-hydrologic-forcings-ml/blob/main/1_fetch/src/get_nwis_data.R#L144)), but a `tar_combine` function may already build in that capability.

jds485:
Should remove the metrics that are the same in both

PR #96, opened 2022-04-23
maps and violin plots of streamflow metrics by cluster

slevin75:
Added a few targets as p5 for the exploratory plots.  This makes a map and violin plot of all the metrics, by cluster.

### DIFF
--- a/5_EDA/src/EDA_metric_plots.R
+++ b/5_EDA/src/EDA_metric_plots.R
@@ -25,19 +25,20 @@ make_EDA_metric_plots <- function(metric,k,cluster_table, metrics_table, gages,o
     metric_quantile <- as.numeric(str_split(metric,pattern="_q")[[1]][2])
     cluster_col<-ifelse(metric_quantile< 0.75,low_quant_cluster_col,high_quant_cluster_col)
   } ##end if
-  
+ 
+   ##renaming the cluster column to make referencing this column easier later on
+  cluster_table <- cluster_table %>% 
+     rename(cluster = all_of(cluster_col))  
   
   ##get the column index of the metric
   metric_col<- which(names(metrics_table)==metric)
   ##renaming the first column to match the cluster and gages ID field
   names(metrics_table)[1] <- "ID"
   
-  
   ##create data frame for plots
   df_plot<- 
-    inner_join(gages[,c("ID","LAT","LON")],cluster_table[,c(1,cluster_col)]) %>%
+    inner_join(gages[,c("ID","LAT","LON")],cluster_table[,c("ID","cluster")]) %>%
     left_join(.,metrics_table[c(1,metric_col)])
-  
 
 
   ##loop through clusters and make a map and violin plot for each cluster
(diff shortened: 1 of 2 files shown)


slevin75:
fhfdc_q0 9_k5_cluster2


The Y-axis label on the map is cut off, but I changed it after I ran the pipeline and I think it should be ok now.  I tested it outside of the pipeline and it looked ok, but I didn't re-run it. So hopefully, it is ok now.

5_EDA/src/EDA_metric_plots.R
```diff
@@ -0,0 +1,78 @@
...
+
+  ##loop through clusters and make a map and violin plot for each cluster
+  states <- map_data("state")
+  fileout_all <- vector()
+  for (cl in 1:k){
+    
+    df<-df_plot %>% 
+      filter(.data[[colnames(df_plot)[5]]]==cl)
+    
+    p_violin<- ggplot(df, aes(x=1,y = .data[[colnames(df)[6]]]))+
```
jds485, line 51: Could we use `cluster_col` and `metric_col` instead of the colnames? Is that what you're expecting to grab here?

5_EDA/src/EDA_metric_plots.R
```diff
@@ -0,0 +1,78 @@
...
+      geom_violin(draw_quantiles = c(0.5))+
+      geom_jitter(height=0,color = "steelblue",alpha=0.5,width=0.2)+
+      xlab(paste("cluster" , cl))
+    
+    p_map<- ggplot(states, aes(x=long, y=lat, group=group)) +
+      geom_polygon(fill="white", colour="gray") +
+      geom_sf(data = df, inherit.aes = FALSE, 
+              aes(color = .data[[colnames(df)[6]]]),
```
jds485, line 59: `metric_col` instead?

5_EDA/src/EDA_metric_plots.R
```diff
@@ -0,0 +1,78 @@
...
+      geom_violin(draw_quantiles = c(0.5))+
+      geom_jitter(height=0,color = "steelblue",alpha=0.5,width=0.2)+
+      xlab(paste("cluster" , cl))
+    
+    p_map<- ggplot(states, aes(x=long, y=lat, group=group)) +
+      geom_polygon(fill="white", colour="gray") +
+      geom_sf(data = df, inherit.aes = FALSE, 
+              aes(color = .data[[colnames(df)[6]]]), 
+              size = 0.75)+
+      scale_color_gradientn(colors=rainbow(5))+
```
jds485, line 61: Colorblind friendly with the scientific rainbow:
`scale_color_scico(palette = 'batlow')`

5_EDA/src/EDA_metric_plots.R
```diff
@@ -0,0 +1,78 @@
...
+      theme(legend.position="bottom",
+            legend.key.size=unit(.5,'cm'))+
+      xlab('Longitude') + 
+      ylab('Latitude') +
+      ggtitle(paste("cluster",cl))
+
+     fname<- paste0(metric,"_k",k,"_cluster",cl,".png")
+     fileout<-file.path(out_dir, fname)  
+     save_plot(filename = fileout, base_height = 5, base_width = 7,
```
jds485, line 70: Did you need to set the height and width for the label to appear? I've never needed to set these to get a good layout

jds485 (changes requested):
Looks good! I have some minor code and plot suggestions

5_EDA/src/EDA_metric_plots.R
```diff
@@ -0,0 +1,78 @@
...
+  fileout_all <- vector()
+  for (cl in 1:k){
+    
+    df<-df_plot %>% 
+      filter(.data[[colnames(df_plot)[5]]]==cl)
+    
+    p_violin<- ggplot(df, aes(x=1,y = .data[[colnames(df)[6]]]))+
+      geom_violin(draw_quantiles = c(0.5))+
+      geom_jitter(height=0,color = "steelblue",alpha=0.5,width=0.2)+
+      xlab(paste("cluster" , cl))
```
jds485, line 54: Remove x-axis ticks
```suggestion
      xlab(paste("cluster" , cl)) +
      theme(axis.ticks.x = element_blank())
```

_targets.R
```diff
@@ -750,5 +752,27 @@ list(
...
-            )  
+            )  ,
+  
+  ###EDA plots
+  ##list of all the metrics names - for dynamic branching
+  tar_target(p5_all_metrics_names,
+             c(colnames(p1_HIT_metrics)[-1],colnames(p1_FDC_metrics)[-1])),
+  
+  ##combined metrics tables
+  tar_target(p5_all_metrics,
```
jds485, line 763: We could label this as `p2` instead (which will be the phase for metrics). It addresses #29

5_EDA/src/EDA_metric_plots.R
slevin75: actually, it turns out I don't need those base_height and base_width - only the scale in the plot_grid function.  I'm not sure exactly why I put those in, lol. I took them out.

5_EDA/src/EDA_metric_plots.R
slevin75: I have cleaned this up so its easier to follow.  We can't use cluster_col or metric_col here because I merged them all into another data frame but I've renamed the cluster column for easier referencing and the metric column can be referenced by its name.

slevin75:
New plots:

dh1_k5_cluster1

slevin75:
@jds485 ok, I have made those changes, let me know if it looks ok to merge.

jds485 (approved):
Thanks, looks good!

PR #102, opened 2022-04-27
63 avg dam features

cstillwellusgs:


1_fetch/src/get_sb_data.R
```diff
@@ -214,6 +214,26 @@ download_children <-function(sites, sb_table_reduced, dldir, workdir, outdir, ou
   return(return_df)
 }
 
+calc_avg_dams <- function(sb_data) {
+  sb_data <- read_csv(sb_data, show_col_types = FALSE) %>%
+    suppressMessages()
+  dams <- sb_data %>%
+    select(COMID, contains("NDAMS"), contains("STORAGE"), contains("MAJOR")) %>%
```
jds485, line 221:
```suggestion
    select(COMID, contains("NDAMS"), contains("NID"), contains("STORAGE"), contains("MAJOR")) %>%
```

jds485: Not sure if anything else will need to change with this additional variable

jds485 (commented):
4th dam variable

1_fetch/src/get_sb_data.R
jds485: Nevermind, Charlie noted that STORAGE captures both NID_STORAGE and TOTAL_STORAGE

jds485 (approved):
Looks good! It is surprising that some of these minimally altered reaches have >400 upstream dams (up to 20 major dams). I think we should check with Ken and Ecoflows if they used CAT or TOT alteration attributes to screen these reaches

cstillwellusgs:
Was just discussing dams on another project. I'm guessing there was some normalization by drainage area to determine 'altered' (I guess a traditional threshold is 103 acre-feet per square mile). Would definitely be curious to learn the procedure for Gages2.1.

Issue #60, opened 2022-03-01
Add target for a table of ScienceBase date last modified


jds485:
There is a way to check the date that a ScienceBase item was last modified with ​`item_get_fields(<sb_id>, 'provenance')​`.

So, I think we can make a target for a table containing the provenance and other ScienceBase item information, and pass that table into the download target. In the download target, we can re-download only those items whose date changed.

jds485:
#101 adds the SB table. We've decided to move the data download check to a new issue, #105

PR #103, opened 2022-04-28
101 streamline sb

cstillwellusgs:
Cleans up the feature variable download and selection process. New target 'p1_feature_vars_g2' can be considered the "cleaned-up" list of science base variables available for feature engineering. Will probably make changes down the road (weighted averages by period of record for each gage instead of simple long-term averages, maybe combining other variables such as all the physiographic regions, etc.). Also switches from static to dynamic branching on the data downloads and adds the log file to track latest updates on sciencebase.


1_fetch/src/get_sb_data.R
```diff
@@ -206,18 +72,95 @@ download_children <-function(sites, sb_table_reduced, dldir, workdir, outdir, ou
...
+  
+  write_csv(sb_log, file_out)
+  return(file_out)
+  
+}
+
+prep_feature_vars <- function(sb_var_data, sites) {
+  
+  data <- sites %>%
+    select(COMID, LAT, LON, npdes, fwwd, strg, devl, cndp)
```
jds485, line 102: These are all for the total upstream area, so I'm recommending adding the TOT prefix. Also moving the as.numeric here because I think these are all numeric columns.

```suggestion
    select(COMID, LAT, LON, npdes, fwwd, strg, devl, cndp) %>%
    rename(TOT_npdes = npdes, TOT_fwwd = fwwd, TOT_strg = strg, TOT_devl = devl, TOT_cndp = cndp) %>% 
    as.numeric()
```

1_fetch/src/get_sb_data.R
```diff
@@ -206,18 +72,95 @@ download_children <-function(sites, sb_table_reduced, dldir, workdir, outdir, ou
...
+  
+  data$npdes <- as.numeric(data$npdes)
+  data$fwwd <- as.numeric(data$fwwd)
+  data$strg <- as.numeric(data$strg)
+  data$devl <- as.numeric(data$devl)
+  data$cndp <- as.numeric(data$cndp)
+  
+  data <- data %>% 
+    group_by(COMID) %>%
+    summarise_all(mean)
```
jds485, line 120: Is this because there are duplicate COMIDs? Can you add a comment?

1_fetch/src/get_sb_data.R
```diff
@@ -206,18 +72,95 @@ download_children <-function(sites, sb_table_reduced, dldir, workdir, outdir, ou
...
+      group_by(COMID) %>%
+      summarise_all(mean)
+    data <- left_join(data, data_temp, by = "COMID")
+  }
+  
+  data$npdes <- as.numeric(data$npdes)
+  data$fwwd <- as.numeric(data$fwwd)
+  data$strg <- as.numeric(data$strg)
+  data$devl <- as.numeric(data$devl)
+  data$cndp <- as.numeric(data$cndp)
```
jds485, line 116: If you accept edit above to move as.numeric
```suggestion
```

1_fetch/src/get_sb_data.R
```diff
@@ -272,5 +205,28 @@ calc_avg_wildfire <- function(sb_data) {
...
     select(COMID, label, avg_annual) %>%
     pivot_wider(names_from = "label", values_from = "avg_annual")
-  return(wildfire)
-}
\ No newline at end of file
+  
+  data <- data %>%
+    select(-ends_with(".y"), -ID, -starts_with("..."), -contains("_S1"), -contains("PHYSIO_AREA"), 
+           -contains("SOHL"), -contains("NDAMS"), -contains("STORAGE"), 
+           -contains("MAJOR"), -contains("TAV"), -contains("PPT"), -contains("WILDFIRE")) %>%
```
jds485, line 212: There are a few TAV and PPT columns we'll want to retain (min, max, average annual). I think filtering with `_TAV_` and `_PPT_` instead works to only remove the monthly TAV and PPT columns

1_fetch/src/get_sb_data.R
```diff
@@ -206,18 +72,95 @@ download_children <-function(sites, sb_table_reduced, dldir, workdir, outdir, ou
...
+    log_item <- tibble(sb_id, last_update)
+    sb_log <- bind_rows(sb_log, log_item)
+  }
+  
+  write_csv(sb_log, file_out)
+  return(file_out)
+  
+}
+
+prep_feature_vars <- function(sb_var_data, sites) {
```
jds485, line 99: Can you comment what columns are expected to exist in both of these variables? That will help us when we want to apply this to ungaged reaches

1_fetch/src/get_sb_data.R
```diff
@@ -206,18 +72,95 @@ download_children <-function(sites, sb_table_reduced, dldir, workdir, outdir, ou
...
+    subset(!(COMID %in% comid_out_conus))
+  
+  land_cover <- land_cover %>%
+    group_by(COMID, unit, new_class) %>%
+    summarise(value = mean(value), .groups = "drop") %>%
+    mutate(label = paste0(unit, "_SOHL_", new_class)) %>%
+    select(COMID, label, value) %>%
+    pivot_wider(names_from = "label", values_from = "value")
+  land_cover <- land_cover %>%
+    subset(!(COMID %in% comid_out_conus))
```
jds485, line 161: Did you rescale the proportions to 100 after filtering? I don't see that step. I think it would go here. You can apply the rescaling to all land cover columns because they all do not sum exactly to 100.

1_fetch/src/get_sb_data.R
```diff
@@ -272,5 +205,28 @@ calc_avg_wildfire <- function(sb_data) {
...
+  data <- data %>%
+    select(-ends_with(".y"), -ID, -starts_with("..."), -contains("_S1"), -contains("PHYSIO_AREA"), 
+           -contains("SOHL"), -contains("NDAMS"), -contains("STORAGE"), 
+           -contains("MAJOR"), -contains("TAV"), -contains("PPT"), -contains("WILDFIRE")) %>%
+    left_join(land_cover) %>%
+    left_join(dams) %>%
+    left_join(weather) %>%
+    left_join(wildfire)
+  
+  rename_dup_headers <- list()
```
jds485, line 218: What causes these duplicates? Could we instead `select(-colname)` in the `left_join`s?

1_fetch/src/get_sb_data.R
```diff
@@ -206,18 +72,95 @@ download_children <-function(sites, sb_table_reduced, dldir, workdir, outdir, ou
...
   write_csv(itemfails, filepath2)
   return_df = c(filepath1, filepath2)
   return(return_df)
 }
 
-calc_avg_dams <- function(sb_data) {
-  sb_data <- read_csv(sb_data, show_col_types = FALSE) %>%
-    suppressMessages()
-  dams <- sb_data %>%
+get_sb_data_log_g2 <- function(sb_var_ids, file_out) {
```
jds485, line 83: Is this specific to g2? I think it also applies to the ungaged sites, so I'm suggesting removing g2. Will need to update name elsewhere, too.
```suggestion
get_sb_data_log <- function(sb_var_ids, file_out) {
```

jds485 (changes requested):
This looks good! I have some suggested changes, and a few comments for rescaling the land cover proportions to 100 after filtering, and checking the SB date last modified before downloading (that may be better for a new PR)

1_fetch/in/sb_var_ids.csv
```diff
@@ -0,0 +1,40 @@
+sciencebase_id,name
```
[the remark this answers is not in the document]
cstillwellusgs: I made it. Personally, I think it is much easier to simply change this .csv if we want to add/remove variables as opposed to pulling in all variables, cutting some, then adding some others back in (all via code and targets). If you disagree we could revert back.

1_fetch/src/get_sb_data.R
```diff
@@ -1,177 +1,42 @@
...
-        print("adding one to dllist")
-        dllist <- c(dllist, idN)
-      }
-      count <- count + 1
-    }
-  }
-  
-  #this gets rid of our first "blank" entry
-  dllist <- dllist[-1]
+get_sb_data <- function(sites, sb_var_ids, dldir, workdir, outdir, out_file_name) {
```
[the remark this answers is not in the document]
cstillwellusgs: I was thinking of that for our other log files as well - nothing depends on them, so it would be possible to miss changes to the log files. I think it should be a separate issue.

1_fetch/src/get_sb_data.R
```diff
@@ -1,177 +1,42 @@
...
+                      list.files(workdir, ".csv", recursive = TRUE, full.names = TRUE, include.dirs = TRUE), 
+                      list.files(workdir, ".TXT", recursive = TRUE, full.names = TRUE, include.dirs = TRUE), 
+                      list.files(workdir, ".CSV", recursive = TRUE, full.names = TRUE, include.dirs = TRUE))
     for (filem in filestomerge)
     {
       print(filem)
       success = "yes"
-      tryCatch(tempfile <- read_delim(filem, delim = ",", show_col_types = FALSE), error = function(e) {success <<- "no"})
+      tryCatch(tempfile <- read_delim(filem, delim = ",", show_col_types = FALSE), 
+               error = function(e) {success <<- "no"})
```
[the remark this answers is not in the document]
cstillwellusgs: That's a questions for @ajsekell

1_fetch/src/get_sb_data.R
cstillwellusgs: If we start renaming feature variables we might want to create a data definition file. With hundreds of different features it is hard to keep track, especially if we begin renaming. Potentially separate issue?

1_fetch/src/get_sb_data.R
cstillwellusgs: Correct. All the feature variable values should be the same (I skimmed, but didn't formally check - that would be easy to do). So here, mean of all simply eliminates the duplicates. I will verify and comment.

1_fetch/src/get_sb_data.R
cstillwellusgs: Good catch!

1_fetch/src/get_sb_data.R
cstillwellusgs: I forgot - good call. Will add in a scaling step. 

Question for self - do other feature variables require rescaling as well??

1_fetch/src/get_sb_data.R
cstillwellusgs: There is one set of feature variables in Science Base that is duplicated (https://www.sciencebase.gov/catalog/item/57976a0ce4b021cadec97890?community=National+Water-Quality+Assessment+-+Hydrologic+Systems+Team+%28NAWQA+-+HST%29) - you will notice there is a FTYPE_PCT_CONUS.txt in addition to FTYPE_PCT_CONUS.zip (which also contains the same .txt file). Rather than hard-code in Andrew's downloading functions to skip over this duplicate, I think it is easier for us to clean up after. Further, if any other feature variables groups encounter this same problem, the code above should work to remove all the doubles (appended with ".y") and rename the originals (remove the appendage ".x").

1_fetch/src/get_sb_data.R
cstillwellusgs: Nope - good catch!

cstillwellusgs:
I don't want to dive back in since I'm heading out in a few minutes. Suggestions noted and I will commit changes early next week. Again, if waiting on me creates a bottle-neck, feel free to make changes yourself.

1_fetch/in/sb_var_ids.csv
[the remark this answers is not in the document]
jds485: Yes, it's definitely easier to work from SB IDs than the xlsx table of variables

1_fetch/src/get_sb_data.R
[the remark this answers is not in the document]
jds485: Okay, sounds good. Part of issue #60 included this download check. I'll add a comment on that issue that we're moving the check to a new issue

1_fetch/src/get_sb_data.R
jds485: Yes, good idea to have a definition file. I think a lot of these are already defined in the vars of interest xlsx file, so that's a good starting point.


These 5 variables are the only ones in the gages dataset. We could rename them in the gages xlsx file instead of renaming them here.

1_fetch/src/get_sb_data.R
jds485: Okay, thanks for that explanation. I don't see where `remove all the doubles (appended with ".y")` happens. It looks like only the `.x` are renamed and `.y` would be kept

1_fetch/src/get_sb_data.R
jds485: There are some other features that could be reduced here. Also happy to do these in a separate step. There are some Nitrogen inputs, NLCD land cover (or maybe you already removed that SB link?) 

# Nitrogen
'NO10AVE', 'NO200AVE', 'NO4AVE', 
# Land cover attributes that are 1 yr, few years, or NLCD
'NLCD', 'IMPV',  'ESTUARY',  'CNPY', 'WRSE'
# physiographic area coverage is not useful - percent of catchment covered by physiographic regions.
'PHYSIO_AREA',
# this runoff is by a coarse HUC level, not reach level. We have a different water balance model runoff that's by reach, so we can drop this one.
'RUN7100'
# Catchment soils - these have NA that are usually water, but somehow the TOT/ACC attributes do not have NAs, which makes me suspicious of the CAT NAs. I'm opting to remove them and use TOT or ACC. 
"CAT_HGA", "CAT_HGAC", "CAT_HGAD", "CAT_HGB", "CAT_HGBC", "CAT_HGBD", "CAT_HGC", "CAT_HGCD", "CAT_HGD", "CAT_HGVAR"

1_fetch/src/get_sb_data.R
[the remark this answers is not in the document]
cstillwellusgs: I think the `<<-` is correct in this case. It searches outside of the immediate brackets for a value named "success" (which was defined a few lines earlier) and re-assigns the new value even though it is outside the local environment.

1_fetch/src/get_sb_data.R
cstillwellusgs: > Nitrogen
'NO10AVE', 'NO200AVE', 'NO4AVE',

These are in the [STATSGO Soil Characteristics, Layer Attributes](https://www.sciencebase.gov/catalog/item/5728decfe4b0b13d3918a9aa?community=National+Water-Quality+Assessment+-+Hydrologic+Systems+Team+%28NAWQA+-+HST%29). Probably will be highly correlated to other soil attributes, but that is probably best to be determined in your feature engineering steps. 

> Land cover attributes that are 1 yr, few years, or NLCD
'NLCD', 'IMPV', 'ESTUARY', 'CNPY', 'WRSE'

Already removed from sb_var_ids.csv - we can add these back in as needed if we decide to use NLCD later. 

> physiographic area coverage is not useful - percent of catchment covered by physiographic regions.
'PHYSIO_AREA',

Already included in line 210. 

> this runoff is by a coarse HUC level, not reach level. We have a different water balance model runoff that's by reach, so we can drop this one.
'RUN7100'

Already removed from sb_var_ids.csv.

> Catchment soils - these have NA that are usually water, but somehow the TOT/ACC attributes do not have NAs, which makes me suspicious of the CAT NAs. I'm opting to remove them and use TOT or ACC.
"CAT_HGA", "CAT_HGAC", "CAT_HGAD", "CAT_HGB", "CAT_HGBC", "CAT_HGBD", "CAT_HGC", "CAT_HGCD", "CAT_HGD", "CAT_HGVAR"

Do you think it would be better to remove these in your later feature engineering steps? I think all of the decisions regarding selections between CAT / ACC / TOT are better suited for the model prep and feature selection stage (even if we know up-front that they might not be useful).

jds485: Yes, the remaining attributes can be removed in a separate target for feature selection

1_fetch/src/get_sb_data.R
cstillwellusgs: This is a good idea - going through this exercise made me realize that the function requires these variables from the Gages2.1 dataset, which will not be available everywhere: npdes, fwwd, strg, devl, cndp. What are we going to do about that, just drop these from our features list?

jds485: I'm keeping those in for modeling for now, but if they're not available everywhere (ungaged locations) we would need to drop them. I think there are some overlaps with the NHD attributes, but they are not exactly the same.

1_fetch/src/get_sb_data.R
cstillwellusgs: `select(-ends_with(".y")`  before the left joins removes the .y, leaving just the .x to be renamed.

1_fetch/src/get_sb_data.R
cstillwellusgs: Added an argument to list out additional variables (columns) to keep from the sites argument (in this case, Gages2.1).

_targets.R
```diff
@@ -284,70 +281,46 @@ list(
...
-             calc_avg_dams(sb_data = p1_sb_data_g2_csv[file_ind])
-             }, 
-             deployment = 'main'
+  
+  ##generate log file to track updates to sb variables
+  tar_target(p1_sb_data_g2_log,
+             get_sb_data_log_g2(sb_var_ids = p1_sb_var_ids,
+                                file_out = "1_fetch/out/logs/sb_update_log.csv"),
+             deployment = "main",
+             format = "file"
```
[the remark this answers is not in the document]
cstillwellusgs: I got an error when adding cue = "always" so I removed - this will be addressed in the other issue about logs

cstillwellusgs:
Ready for your review. I 'resolved' a lot of your suggestions to clear up my screen but please feel free to unresolve any as you see fit.

jds485 (approved):
Looks good! I have a few small suggestions, and you don't have to accept them. You can merge after addressing those

_targets.R
```diff
@@ -284,70 +281,48 @@ list(
...
-                               outdir = "./1_fetch/out/sb",
-                               out_file_name = "sb_data_g2.csv"),
+             get_sb_data(sites = p1_sites_g2, 
+                         sb_var_ids = p1_sb_var_ids,
+                         dldir = "./1_fetch/out/sb/dldir", 
+                         workdir = "./1_fetch/out/sb/workdir",
+                         outdir = "./1_fetch/out/sb/data",
+                         out_file_name = "sb_data_g2_"),
+             map(p1_sb_var_ids),
+             iteration = "list",
```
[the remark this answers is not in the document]
cstillwellusgs: Feel free to change this up later, this was how I was able to figure it out haha.

PR #108, opened 2022-05-03
Address messages in p1_feature_vars_g2

cstillwellusgs:
Line 142 of get_sb_data.R converts the GAGE_ID column back into character. Many (but not all) of these sites have leading zeros. Keep an eye out, and if you have an idea on how to retain those zeros let me know. I'll think on it.

### DIFF
--- a/1_fetch/src/get_sb_data.R
+++ b/1_fetch/src/get_sb_data.R
@@ -131,9 +131,16 @@ prep_feature_vars <- function(sb_var_data, sites, retain_vars) {
   #'@return data frame with COMID column appended by all feature variables of interest; 
   #'time-varying features converted to long-term averages where applicable
   
+  options(dplyr.summarise.inform = FALSE)
+  
   data <- sites %>%
     select(COMID, all_of(retain_vars)) %>%
     mutate(across(where(is.character), as.numeric))
+  
+  if("ID" %in% retain_vars) {
+    data <- rename(data, GAGES_ID = ID) %>%
+      mutate(GAGES_ID = as.character(GAGES_ID))
+  }
   
   #join all sciencebase data to single data frame with comids
   for (i in 1:length(sb_var_data)) {
@@ -186,6 +193,7 @@ prep_feature_vars <- function(sb_var_data, sites, retain_vars) {
     subset(!(COMID %in% comid_out_conus)) %>%
     group_by(COMID, unit, year, new_class) %>%
     mutate(lc_adj = (value / lc_sum) * 100) %>%
+    ungroup() %>%
     select(COMID, unit, new_class, lc_adj) %>%
     group_by(COMID, unit, new_class) %>%
     summarise(value = mean(lc_adj), .groups = "drop") %>%
--- a/1_fetch/src/get_sb_data.R
+++ b/1_fetch/src/get_sb_data.R
@@ -135,7 +135,7 @@ prep_feature_vars <- function(sb_var_data, sites, retain_vars) {
   
   data <- sites %>%
     select(COMID, all_of(retain_vars)) %>%
-    mutate(across(where(is.character), as.numeric))
+    mutate(across(where(is.character)  & !starts_with('ID'), as.numeric))
   
   if("ID" %in% retain_vars) {
     data <- rename(data, GAGES_ID = ID) %>%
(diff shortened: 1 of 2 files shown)


1_fetch/src/get_sb_data.R
```diff
@@ -131,10 +131,17 @@ prep_feature_vars <- function(sb_var_data, sites, retain_vars) {
   #'@return data frame with COMID column appended by all feature variables of interest; 
   #'time-varying features converted to long-term averages where applicable
   
+  options(dplyr.summarise.inform = FALSE)
+  
   data <- sites %>%
     select(COMID, all_of(retain_vars)) %>%
     mutate(across(where(is.character), as.numeric))
```
jds485, line 138: I think this will work. Then the mutate in the if statement for GAGES_ID is not needed
```suggestion
    mutate(across(where(is.character)  & !starts_with('ID'), as.numeric))
```

1_fetch/src/get_sb_data.R
```diff
@@ -131,10 +131,17 @@ prep_feature_vars <- function(sb_var_data, sites, retain_vars) {
...
   
+  if("ID" %in% retain_vars) {
+    data <- rename(data, GAGES_ID = ID) %>%
+      mutate(GAGES_ID = as.character(GAGES_ID))
+  }
+  
   #join all sciencebase data to single data frame with comids
   for (i in 1:length(sb_var_data)) {
     filepath <- sb_var_data[[i]][1]
```
jds485, line 147: This will also have to be edited because `summarise_all` will try to summarize the GAGES_ID. I'm not sure of a general way to handle this problem, but a quick fix is to only do this step `if(!("ID" %in% retain_vars))`
```
 #handles any duplicated COMIDs
  data <- data %>% 
    group_by(COMID) %>%
    summarise_all(mean)
```

1_fetch/src/get_sb_data.R
```diff
@@ -186,6 +193,7 @@ prep_feature_vars <- function(sb_var_data, sites, retain_vars) {
     subset(!(COMID %in% comid_out_conus)) %>%
     group_by(COMID, unit, year, new_class) %>%
     mutate(lc_adj = (value / lc_sum) * 100) %>%
+    ungroup() %>%
     select(COMID, unit, new_class, lc_adj) %>%
     group_by(COMID, unit, new_class) %>%
     summarise(value = mean(lc_adj), .groups = "drop") %>%
```
jds485, line 199: Adding comment on the nearest line that I can comment on.

Can you add `by = COMID` to these joins to remove the message that it's joining by COMID?
    left_join(land_cover) %>%
    left_join(dams) %>%
    left_join(weather) %>%
    left_join(wildfire)

jds485 (changes requested):
I have a few suggestions for how to handle the leading 0s + some edits that I think are needed because we're keeping a character column that can't be summarized

cstillwellusgs:
Removing `options(dplyr.summarise.inform = FALSE)` means you will still get the message "summarize() grouped by: ..."

1_fetch/src/get_sb_data.R
```diff
@@ -145,34 +149,62 @@ prep_feature_vars <- function(sb_var_data, sites, retain_vars) {
...
-                                 class == 9 ~ "FOREST", class == 10 ~ "FOREST", 
-                                 class == 11 ~ "GRASSLAND", class == 12 ~ "SHRUBLAND", 
-                                 class == 13 ~ "CROPLAND", class == 14 ~ "HAY/PASTURE", 
-                                 class == 15 ~ "WETLAND", class == 16 ~ "WETLAND", 
-                                 class == 17 ~ "ICE/SNOW")) %>%
-    group_by(COMID, unit, year, new_class) %>%
-    summarise(value = sum(value)) %>%
-    mutate(lc_sum = sum(value)) %>%
-    ungroup()
+  if("ID" %in% retain_vars) {
```
jds485, line 159: Did you check that these ifs for land cover, fire, etc. are needed? I'd expect the resulting tables by COMID to be the same as the tables by GAGES_ID because the data are available for COMIDs.

cstillwellusgs: There are more GAGES_ID values than COMID values - without the if statement, the resulting data frames have a different number of rows and mess with the left_join() at the end.

jds485 (approved):

1_fetch/src/get_sb_data.R
cstillwellusgs: yeah. that's what i was expecting too. let me play around with an inner join or something, maybe that is what i need

PR #117, opened 2022-06-29
Quick PR to get main up to date with prediction models branch

jds485:
This PR updates the main branch with 2 small commits that should have been included in #110


jds485:
@cstillwellusgs: I think this code difference is why you were seeing outdated targets when you tried to copy the _targets and */out/* folders.

PR #119, opened 2022-07-07
118 dominant region

### LINKED ISSUE #118: Convert separate physiographic features to single dominant physiographic region

cstillwellusgs:
Closes #118


1_fetch/src/get_sb_data.R
```diff
@@ -258,11 +258,31 @@ prep_feature_vars <- function(sb_var_data, sites, retain_vars) {
...
+  phys_region <- data %>%
+    select(COMID, contains("_PHYSIO_")) %>%
+    select(-contains("AREA")) %>%
+    group_by(COMID) %>%
+    summarise_all(mean) %>%
+    pivot_longer(!COMID, names_to = "name", values_to = "value") %>%
+    mutate(unit = str_sub(name, 1, 3), 
+           region = str_sub(name, 12)) %>%
+    group_by(COMID, unit) %>%
+    summarise(dom_phys_reg_ind = which.max(value), .groups = "drop") %>%
```
jds485, line 271: Have you handled the case when more than one region has the same maximum coverage area?

cstillwellusgs: I have not... it appears that `which.max()` (and all `which()` functions) default to the smallest index when there are ties. Not sure how we would choose between one or the other in our case, will hold off for today's meeting.

1_fetch/src/get_sb_data.R
```diff
@@ -258,11 +258,31 @@ prep_feature_vars <- function(sb_var_data, sites, retain_vars) {
...
+    mutate(dom_phys_reg = if_else(dom_phys_reg_ind < 3,
```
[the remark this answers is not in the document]
cstillwellusgs: Editing my code to handle this in a more transparent way. You should see in upcoming Commit. 

There are 25 regions, but they start at 0 (no region), skip 2 for some reason... so this was to correct for that. Now using `which.max()` in a better way to return the full row (with region column) instead of just the index (which were off my 1 for regions 0 and 1).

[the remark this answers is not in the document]
cstillwellusgs: https://www.sciencebase.gov/catalog/file/get/57dafd3ae4b090824ffc32f1?f=__disk__9b%2Fe6%2F04%2F9be604f55425b2691158260123a914cd0efae0da&transform=1&community=National+Water-Quality+Assessment+-+Hydrologic+Systems+Team+%28NAWQA+-+HST%29&allowOpen=true

1_fetch/src/get_sb_data.R
cstillwellusgs: We ran out of time to discuss this today. There aren't any ties in Gages2.1 so it isn't technically an issue for the moment, but there could be COMIDs in our prediction space with ties. I'm thinking if there were to be a tie in either `ACC_` or `TOT_` regions, we could use the dominant `CAT_` region as the tiebreaker (defer to the physical location of the reach rather than upstream watershed area). I would be surprised if there were any ties in `CAT_` but I suppose it is possible... thoughts?

jds485: I agree it would be less likely for CAT to have ties. That sounds like a good approach to me

cstillwellusgs:
@jds485 addressed your comments, ready for your review.

1_fetch/src/get_sb_data.R
```diff
@@ -258,11 +258,54 @@ prep_feature_vars <- function(sb_var_data, sites, retain_vars) {
...
+  for (i in unique(phys_region$COMID)) {
+    phys_reg_comid <- filter(phys_region, COMID == i)
+    
+    dom_phys_reg_comid <- tibble()
+    for (j in unique(phys_reg_comid$unit)) {
+      phys_reg_comid_unit_ties <- filter(phys_reg_comid, unit == j) %>%
+        arrange(desc(value))
+      
+      #if tie for dominant phyiographic region, default to CAT value
+      if (phys_reg_comid_unit_ties[[1,3]] > phys_reg_comid_unit_ties[[2,3]]) {
```
jds485, line 281: Should this line have single brackets `[]` instead of double `[[]]`? All other changes look good to me! I think merging should wait until after #122 is merged so that you can transfer over the _targets data with the script in that PR

cstillwellusgs: Either single or double brackets work, I preferred double in this instance because it just returns a numeric value rather than the number with the label. 

I'll review #122 shortly.

PR #124, opened 2022-07-11
Split data wtih spatial nestedness

slevin75:
This PR includes the following updates:

-   p4_nested_groups uses the add_nested_group_id function to determines which gages in the nested_gages matrix are nested based on the nested_threshold.  Then it assigns a unique nested_group_id to every un-nested gage or groups of nested gages. 
- split_data function was updated to ensure that any gage with the same nested_group_id are not split between training and testing datasets.  If any nested gages were split up, it takes the gages in the testing dataset and adds them to the training dataset, then finds an equal number of un-nested gages in the training dataset to replace the moved gages in the testing data set. 
- post_process_cvfold : this function post processes the vfold_cv function to ensure that any nested gages are not separated between the in / out of sample data using the same process as for the split_data function.
- Updated the Boruta_screening function calls and train_models_grid to work with new functions.


.gitignore
```diff
@@ -61,3 +61,6 @@ vignettes/*.pdf
 
 # R Environment Variables
 .Renviron
+
+#slurm files on hpc
+/slurm/
```
jds485, line 66: I have been git tracking these files. It may not make sense to track the tar_make scripts because they're short, but the #122 scripts we'll want to have tracked. Can you remove these from the .gitignore?

6_predict/src/train_models.R
```diff
@@ -16,11 +18,50 @@ split_data <- function(data, train_prop){
...
+  
+  ##find any nested group ids that are in both data sets
+  intersection_ids<-intersect(training$nested_group_id, testing$nested_group_id)
+  
+  ##find all un-nested group ids
+  unnested<- nested_groups %>%
+    group_by(nested_group_id) %>%
+    count()%>%
+    filter(n ==1)%>%
+    select(nested_group_id)
```
jds485, line 33: Format edit
```suggestion
  intersection_ids <- intersect(training$nested_group_id, testing$nested_group_id)
  
  ##find all un-nested group ids
  unnested <- nested_groups %>%
    group_by(nested_group_id) %>%
    count() %>%
    filter(n == 1)%>%
    select(nested_group_id)
```

4_setup_crossval/src/cross_validation_functions.R
```diff
@@ -1,15 +1,10 @@
 
-get_nested_gages<-function(gagesii,nav_distance_km){
-  gage_IDs <- as.character(gagesii$ID)
-  COMIDs <- as.character(gagesii$COMID)
+get_nested_gages<-function(gagesii,nav_distance_km, screened_site_list){
```
jds485, line 2: Can you edit the description using this roxygen format for this function and the following?

  #' @description 
  #'
  #' @param
  #' @return

jds485 (changes requested):
Thanks, Sara! This looks good overall. I have some minor comments to address.

Did you run the pipeline after making these changes? I'm curious how the p6 performance metrics look using a 50% threshold vs. the 0% threshold in our HydroML presentation

6_predict/src/train_models.R
```diff
@@ -16,11 +18,50 @@ split_data <- function(data, train_prop){
...
+  
+  ##identify nested gages to move from testing to training and
+  ##unnested replacements to move from training to testing
+  move_gages<- testing %>%
+    filter(nested_group_id %in% intersection_ids)
+  
+  replacement_gages<-training %>%
+    filter(nested_group_id %in% unnested$nested_group_id)%>%
+    sample_n(size = nrow(move_gages))
```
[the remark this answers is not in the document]
slevin75: So my initial thought on this was that it was better to have the nested basins in the training group because I feel like having potential duplicates in the testing set, which is smaller, might cause more bias than in the training group.  So I moved all the nested basins that got split up to the training data.  But, if we want to keep that initial split proportion of nested basins the same, maybe I should only replace them with other nested basins?

.gitignore
jds485: I just looked at the files on Caldera and updated to get the correct plot

slevin75: oh, oops!

jds485: No problem! The names of those files are not the most intuitive

jds485: The result is about the same. Do you know how many nested gages are in each region? That could be a helpful number to have for each region

6_predict/src/train_models.R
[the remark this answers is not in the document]
jds485: By duplicates, do you mean nested gages?

That is a good point about trying to keep the test set free of nested gages. It could be challenging to force the proportion to be the same because it might be that it's not possible to do that. I'm okay with leaving it as-is for now. I think the larger problem (not for us) is how to consider in the loss function that nested gages are not independently contributing to the loss

.gitignore
jds485: The regions are in the `cluster_table`. There are different regions for high and med-high, so a total of 10 regions. 

> With a nestedness threshold of 0.5, there are a total of only 91 nested gages in the whole CONUS and there are no large groups of nested gages.

That probably explains why the results were so similar between our runs.

> we can test the model results with a variety of thresholds, but really we want to find a meaningful threshold in terms of how similar the basins are, right?

Yes, that's right. I think it makes sense. What would we use as the similarity metric?

slevin75: I'm not sure really - maybe the mean pearson correlation between all the pairs of streamflows in a nested group, or else a euclidean distance between basin characteristics or flow metrics? I don't know, we might have to talk about it. I wonder if this has been done before? I feel like someone has probably done this.

jds485: @FlowChar any thoughts on this similarity metric to select a threshold for nestedness, Ken?

slevin75: > The result is about the same. Do you know how many nested gages are in each region? That could be a helpful number to have for each region

Here is a table with the total number of gages, number of distinct nested groups and number of nested gages in each region for a nestedness_threshold of 0.5.  While I was doing this, I noticed there are a number of nested groups that are split between different cluster regions, in which case they were not included in this table, but is interesting because we are assuming that nestedness means that they are similar enough that they would not be considered independent but they were apparently not similar enough to be placed in the same clusters. 

   flow_region cluster totaln nested_groups nested_gages
     midhigh       1    203             9                   18
      midhigh      2    308            13                  28
      midhigh      3    638            12                  24
      midhigh      4     89             4                     8
      midhigh      5    141             1                    2
         high         1    183             6                  12
         high         2    752            28                 58
         high         3    102             4                   8
         high         4    172             2                   4
        high          5    170             1                   2

FlowChar: An idea on how to measure how similar a pair of gages are to each other in terms of flow is to normalize all monthly flows by drainage area (highs, lows, and everything in between should be captured) then do a simple stats test between the two distributions to see if they are pulled from the same population. Every pair of gages will have an associated DA overlap percentage. Bin the DA percentages and within each bin show the proportion of pairs that pass/reject the null hypothesis. Choose the bin as the threshold that has the largest ratio of rejections of the null hypothesis. I am suggesting a monthly time step so we get away from the 'lagging' problem of trying to compare flows at two different locations on the same day. (Note: for the GAGES 2.1 REF gages, I have all the monthly values already calculated for all years of records that can be shared).

jds485: > Here is a table...

Thanks! I think this would be useful to keep as a target (e.g., write the table to a csv file).

It seems like 2-3 nested gages is most common, so that explains why we do not see much of a difference for our model results.

> there are a number of nested groups that are split between different cluster regions...

That is interesting! These could be gages on the boundary between classification into one cluster or another. Did you map out where these were located? It's also possible that the flow generation process in headwaters is different than events further downstream.

jds485: @slevin75: do you want to try doing what Ken suggested to evaluate a nestedness threshold?

_targets.R
```diff
@@ -103,6 +104,8 @@ window_length <- 20  ##needs to be <= complete_years
 increment <- 1
 min_yrs_in_window<- 15  ##minimum number of years of data required within a window
 min_windows <- 10  ##Must have this many windows available in order to plot 
+###percentage of drainage area overlap between nested basins above which they will be grouped together
+nested_threshold<-.5
```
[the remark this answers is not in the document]
jds485: Okay, sounds good! I think that makes sense. And because we have few gages that are nested within each region, I am not concerned about testing different thresholds right now. I think our time can be better used working on other tasks

jds485 (approved):
Looks good! I replied to some of your comments, but I think this is ready to merge. I think it will help to have PR #122 merged first because that has scripts that you can use to transfer the _targets data from your fork to USGS main

PR #125, opened 2022-07-19
Low flow metrics and regions

jds485:
The PR computes low flow metrics and regions using a similar method as is used for high flow metrics and regions.

Low flow metrics for each flow magnitude include: duration, frequency, volume, and minimum flow for events less than or equal to the flow threshold, and the volume for events greater than the threshold.
A set of targets is also made for those same metrics except the volume for events greater than the threshold.

There are sites whose quantiles are 0 flow (intermittent streams). For these sites, the seasonal (fractional) flow volume less than the threshold is computed as the number of events per season instead of flow volume per season. The minimum flow fraction is assigned equally to all seasons in which 0 flows occur.

All metrics are rounded to 3 decimal places. Some sites have quantile magnitudes that are separated by < 0.001, so the tabulated magnitude metric has the same value for several quantiles. The other flow metrics were computed from the un-rounded magnitude values.

I think we can retain all sites with 0 flows for clustering as a result of their other metrics being non-zero. It could actually help to know if a site has several quantiles with the same value because it indicates the intermittency of the flows at that site.

Edits:
Calc_FDC - added the `threshold_type` argument to specify if metrics are to be computed for low (below) or high (above) the specified threshold. Other edits address quirks to handle low flows, and function and parameter descriptions.
seasonal_metric_cluster - added fixed = TRUE to grep so that 0.005_ would not be grabbed when using 0.5_ as the pattern. Also added function and parameter descriptions.
EDA_metric_plots - I added 3 arguments to generalize the function to work for low flow quantiles. Also added function and parameter descriptions, and mode format edits.
Calc_HIT - added function and parameter descriptions


jds485:
After these edits, the p6 predict targets were all not outdated, so that confirms that the edits I made to functions to handle low flows did not affect the high flow metric values.

It seems like as a result of random seeds between Charlie's run (on which this PR is built) and Sara's run (most recent previous predict commit) we now see different results for the testing in rain region (training for all of CONUS is better than training in only rainfall dominated). That reinforces a known gap of needing to run several replicates (random seeds) for each predict target and compare their RMSEs considering the sampling uncertainty.

PR #127, opened 2022-07-28
created function to map residuals and variogram of residuals

slevin75:
Included in this pr:

I changed the predict_test_data function to output the gage_ids where the predicted and observed values are so that I could link it to the gagesii LAT LONs.  

The make_residual_map creates the map of residuals (obs - predicted) and maps them along with a variogram plot.  I wasn't sure if I should standardize the residuals in any way, so for now I haven't.  Can change this if needed.  

I made maps corresponding to each of the pred_obs_scatterplots in the out/pred_obs folder.

### DIFF
--- a/6_predict/src/plot_diagnostics.R
+++ b/6_predict/src/plot_diagnostics.R
@@ -306,7 +306,7 @@ make_residual_map <- function(df_pred_obs, sites_g2_sf, metric, pred_gage_ids, r
     ggtitle(paste("metric= ",metric,"    region = ", region ))
   
   p2 <- plot(variogram(resid ~1, df,
-                       cutoff = 100, 
+                       cutoff = 1000, 
                        width = 2,
                        cressie = TRUE),
              asp=1)
--- a/6_predict/src/plot_diagnostics.R
+++ b/6_predict/src/plot_diagnostics.R
@@ -269,12 +269,11 @@ make_residual_map <- function(df_pred_obs, sites_g2_sf, metric, pred_gage_ids, r
   #' @param df_pred_obs is a data frame with observed (obs) and predicted (.pred) from 
   #' the output of  predict_test_data function. if left null and from_predict is true, 
   #' it will predict it from the model_wf (copied these lines from the plot_pred_obs function)
-  #' @param sites_g2_sf  gagesii spatial ata object with LAT LON and ID
+  #' @param sites_g2_sf  gagesii spatial data object with LAT LON and ID
   #' @param metric  metric that is being predicted
   #' @param pred_gage_ids vector of gage ids that correspond to each row of the df_pred_obs
   #' @param region  region of the model/prediction
   #' @param directory where output figures are saved
-  #' 
 
   if(from_predict){
     #predict from provided workflow and data
@@ -290,30 +289,39 @@ make_residual_map <- function(df_pred_obs, sites_g2_sf, metric, pred_gage_ids, r
     mutate(resid = obs - .pred)
   
   states <- map_data("state")
-  
+  limit <- quantile(df$resid, probs = c( 0.1, 0.9))
   p1<-ggplot(states, aes(x=long, y=lat, group=group)) +
-    geom_polygon(fill="white", colour="gray") +
+    geom_polygon(fill="gray60", colour="gray80") +
     geom_sf(data = df, inherit.aes = FALSE, 
             aes(color = .data[["resid"]]), 
-            size = 0.75)+
-    scale_color_scico(palette = 'batlow')+
+            size = 0.5)+
+    scale_color_scico(palette = 'roma',
+                      midpoint = 0,
+                      limits = limit,
+                      oob = scales::squish)+
     theme(legend.position="bottom",
-          legend.key.size=unit(.5,'cm'))+
+          legend.key.size=unit(.75,'cm'))+
     xlab('Longitude') + 
     ylab('Latitude')+
     ggtitle(paste("metric= ",metric,"    region = ", region ))
   
+  p2 <- plot(variogram(resid ~1, df,
+                       cutoff = 100, 
+                       width = 2,
+                       cressie = TRUE),
+             asp=1)
   
-  p2 <- plot(variogram(resid~1, df))
   
   fname<-paste0(out_dir, "/resid_map_", metric, "_", region, ".png")
   
   save_plot(filename = fname, 
             plot = plot_grid(p1,p2, scale = c(1,.8)),
-            base_width = 8,
+            base_width = 10,
             bg="white")
   
+  
 
+  
 
   return(fname)
 }
--- a/6_predict/src/plot_diagnostics.R
+++ b/6_predict/src/plot_diagnostics.R
@@ -261,6 +261,66 @@ plot_pred_obs <- function(df_pred_obs, metric, region, out_dir,
 }
 
 
+make_residual_map <- function(df_pred_obs, sites_g2_sf, metric, pred_gage_ids, region, out_dir,
+                              from_predict = FALSE, model_wf = NULL, pred_data = NULL){
+  #' @description this function creates maps of residuals (obs - predicted) and a
+  #' variogram plot
+  #' 
+  #' @param df_pred_obs is a data frame with observed (obs) and predicted (.pred) from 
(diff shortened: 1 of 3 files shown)


6_predict/src/plot_diagnostics.R
```diff
@@ -261,6 +261,66 @@ plot_pred_obs <- function(df_pred_obs, metric, region, out_dir,
...
+    mutate(resid = obs - .pred)
+  
+  states <- map_data("state")
+  
+  p1<-ggplot(states, aes(x=long, y=lat, group=group)) +
+    geom_polygon(fill="white", colour="gray") +
+    geom_sf(data = df, inherit.aes = FALSE, 
+            aes(color = .data[["resid"]]), 
+            size = 0.75)+
+    scale_color_scico(palette = 'batlow')+
```
jds485, line 299: We should use a diverging colorbar for the residuals (scico has those) and force the center to be 0. We may want to also force that the min and max values for the colorbar are the same.

For residuals, I like to use a 2-color colorbar with different shades of each color

6_predict/src/plot_diagnostics.R
```diff
@@ -261,6 +261,66 @@ plot_pred_obs <- function(df_pred_obs, metric, region, out_dir,
...
+    theme(legend.position="bottom",
+          legend.key.size=unit(.5,'cm'))+
+    xlab('Longitude') + 
+    ylab('Latitude')+
+    ggtitle(paste("metric= ",metric,"    region = ", region ))
+  
+  
+  p2 <- plot(variogram(resid~1, df))
```
jds485, line 307: What are the distance units? km? I think we should increase the xlimit to 100 km and increase the number of lags to compute to 50 or so

There are a few outlier residuals that result in high short-distance semivariance. We can try plotting a 3rd panel with a robust (to outlier point-pairs) semivariance. There is a function to do that within gstat but I don't remember off the top of my head.

6_predict/src/plot_diagnostics.R
```diff
@@ -261,6 +261,66 @@ plot_pred_obs <- function(df_pred_obs, metric, region, out_dir,
...
+  df<- bind_cols(lat_lons,df_pred_obs) %>%
+    mutate(resid = obs - .pred)
+  
+  states <- map_data("state")
+  
+  p1<-ggplot(states, aes(x=long, y=lat, group=group)) +
+    geom_polygon(fill="white", colour="gray") +
+    geom_sf(data = df, inherit.aes = FALSE, 
+            aes(color = .data[["resid"]]), 
+            size = 0.75)+
```
jds485, line 298: we can decrease the point size a little. A lot of points overlap

6_predict/src/plot_diagnostics.R
```diff
@@ -261,6 +261,66 @@ plot_pred_obs <- function(df_pred_obs, metric, region, out_dir,
...
 
+make_residual_map <- function(df_pred_obs, sites_g2_sf, metric, pred_gage_ids, region, out_dir,
+                              from_predict = FALSE, model_wf = NULL, pred_data = NULL){
+  #' @description this function creates maps of residuals (obs - predicted) and a
+  #' variogram plot
+  #' 
+  #' @param df_pred_obs is a data frame with observed (obs) and predicted (.pred) from 
+  #' the output of  predict_test_data function. if left null and from_predict is true, 
+  #' it will predict it from the model_wf (copied these lines from the plot_pred_obs function)
+  #' @param sites_g2_sf  gagesii spatial ata object with LAT LON and ID
```
jds485, line 272:
```suggestion
  #' @param sites_g2_sf  gagesii spatial data object with LAT LON and ID
```

6_predict/src/plot_diagnostics.R
```diff
@@ -261,6 +261,66 @@ plot_pred_obs <- function(df_pred_obs, metric, region, out_dir,
...
+  #' @param metric  metric that is being predicted
+  #' @param pred_gage_ids vector of gage ids that correspond to each row of the df_pred_obs
+  #' @param region  region of the model/prediction
+  #' @param directory where output figures are saved
+  #'
```
jds485, line 277:
```suggestion
  #' 
  #' @return filepath to plot
```

jds485 (approved):
Thanks, this looks good! I have a few comments on plot edits. Can you paste the resulting figure after you've addressed them?

jds485:
Sorry, did not mean to hit approve. I meant to hit request changes

6_predict/src/plot_diagnostics.R
jds485: cressie = TRUE for robust semi-variogram

slevin75: I actually don't really know what I am doing with the variogram.  I think the limit and number of lags are set with cutoff and width, respectively? I'm not sure what the units are though - the x and y are in lat lon, so I guess they are in degrees? Should I project it first so they have units of distance?

jds485: Based on the documentation, you don't need to use a projected coordinate system. You can set `projected = FALSE` and distances will be computed in km units.
Are we currently using a projected or unprojected coordinate system?

Yes, the cutoff will set xmax and width sets the increments between points. So I'd suggest using 100 km for cutoff and 2 km for width.

6_predict/src/plot_diagnostics.R
slevin75: @jds485  I've been playing around with color scales and one problem is that the distribution of the residuals is very long tailed, so when I map it and center it at zero, basically the only thing you see are the extreme outliers and it doesn't seem super helpful:
image

One thing I could do is set the gradient limit to some fixed values and then anything out of those bounds would just get the maximum value color.  Below I set the limits at the 10th and 90th percentiles.  I haven't quite figured out how to change the legend labels so that it shows <=6 and >=6 or something, but I supposed it can be done.  
image


Anyway - let me know what you think or if you have a different way to map them.

jds485: I like the second option better, and that's the method I expected we would need to use. If we use white as 0, then we should make the basemap of the states a dark color (I think the scico package is visually optimal when using black as the background, so could try using black).

slevin75: ok here is an example with the updated maps and variogram.  I tried putting both the variogram and robust variogram on there but it was causing me more time and headaches than I can deal with on a Friday, the scaling of the three plots on 1 figure was just haywire and I tried many solutions and they all failed.  So - if it is important to have both variograms, I can print them out together in a separate figure from the map, otherwise this is what the plots look like.  I think I got all the comments addressed in the latest commit.  If this figure looks ok, I will merge.

image

jds485: Thanks! Did you check if the units are in degrees or km? It looks like it could be degrees based on having larger residuals in the West located about 30 degrees from lower values in the East. 

These plots look good to me.

slevin75: The units are still in lat long but I don't think that is a problem for the map because the lat long units shouldn't affect the residuals.  I'm not exactly sure what the different scenarios are, so I'm not sure why the west does so much poorer here.  

The variogram I think might still be messed up because of the units though.  There is an argument in the variogram function, you can set project=FALSE but when I do this I get an error. So I am thinking maybe I need to project the lat long data prior to the variogram function but I am not sure how to do this.  The examples I've found online are all rather esoteric and I don't really know what I'm doing.

slevin75: ok - I think I have the variogram right now.  I added the  crs to the pq_sites_g2_sf target and this is how the variogram looks now:  I'm pretty sure these are in km now.  

image

jds485: Thanks, that makes sense to me looking at the values we have. I'm not worried about the < 20 km variability here because we have neighboring residuals with large differences in value. A semivariance of 25 corresponds to an average difference of about 7 and that's believable to me.

100 km is about 1 degree. Because our map is 10 degree tick marks, using 1000 km as the max range could be better visually.

slevin75: ok - here it is with a 1000 km max:

image

PR #128, opened 2022-07-29
104 g2.1 duplicates

cstillwellusgs:
- Added number of gages to the 'drop_gages' list (gages in gages 2.1 located on the same comid; overlapping periods of record, I omitted the shorter of the two records available)
- Combined periods of record from three pairs of gages (gages in gages2.1 located on the same comid; non-overlapping periods of record, so they were combined to form one record).

To maintain proper site numbering conventions I had to select one number to move forward with. I'm not sure of the implications down the road (for instance, the residual maps - the "03584000" data actually includes "03584000" and "03584020"). 

I'm updating targets and such on my branch now. Assuming that goes smoothly this should be ready for review.


cstillwellusgs:
Updated all targets and outfiles. However, now realizing that subsequent functions following `prescreen_daily_data()` may have been impacted (peak flows, other screening procedures, etc.). I think it will be best to discuss during Monday's meeting.

cstillwellusgs:
@slevin75 All targets running now (still taking forever... froze overnight??). Go ahead and review what is here, hopefully nothing else needs to change but I will verify once the targets are complete.

slevin75 (approved):
@cstillwellusgs ok - I think this is good to go if you are finished with it.  I stepped through some of the changes to make sure I understood what was going on and it all looked fine to me.

PR #130, opened 2022-08-22
restructured pipeline

slevin75:
In this pr, I restructured the pipeline and put each numeric group of targets in their own file, so the targets file now includes source calls to 1_fetch, 2_flow_metrics, 3_cluster, 4_setup_crossval, 5_EDA, and 6_predict.  I wasn't quite sure where to put the moving window/nonstationarity targets so I just stuck them in with the p4_targets.  Maybe they should go in the 2_flow metrics_group?  I reran the whole pipeline to make sure it all ran.


jds485:
I'm suggesting that Charlie review instead of me because I'm low on hours for this project

cstillwellusgs:
Successfully ran the pipeline (after guidance from Jared). All targets updated. Good to merge!

cstillwellusgs (approved):

PR #136, opened 2022-09-16
Multiclass Region Prediction Model

### LINKED ISSUE #144: Classification model for region membership
Needed to assign ungaged reaches to each cluster region.

Train a classification model to predict class labels (region) of each gage that we used for flow metric cluster analysis.
- classification model using seasonal flow metric method
- classification model using annual flow metric method
- compare classification performance (confusion matrix, F1), and feature importance

jds485:
PR for developing classification models that predict membership of gages to model regions. Uses a probability forest that predicts the conditional probability of a class given attributes. 20 bootstrap datasets are used for model fitting, and the average probability over those 20 models is used for estimation.

The largest function to review is in `train_multiclass_models.R`, which builds the models and saves summary performance information. 

I did not run `tar_make` to build all targets because I am expecting that dependency targets will be updated after the estimated data from that PR are complete. And there are several other open issues for attribute handling that will affect these models, too.

I'm not sure how to download all of NHDPlusV2 on Linux (documentation has instructions for Mac and Windows only) so I'm providing an example of using the predicted probabilities to estimate the most likely cluster region for each gage in our database. Looks very similar to the estimated cluster map, but there are ~1000 more gages than were used for clustering:
LikelyRank1_map

Closes #144


jds485:
Documenting feedback from review discussion with Ken:
- Add function return info
- For the most likely rank plots, do not plot the point / reach if the probability is < specified threshold (e.g., 5%)
- Add model rownames to the saved csvs (Brier, F1, etc.)

jds485:
@FlowChar: I think I addressed your comments. Let me know if this is good to merge

6_predict/src/train_multiclass_models.R
```diff
@@ -0,0 +1,473 @@
...
+      ###################################################################
+      #MODEL TRAINING/VALIDATION
+      ###################################################################
+      #Random Forest-RF
+      ###################################################################
+      #Running Initial RF model to identify top features based on importance
+      #uses default mtry value
+      RF_init <- ranger::ranger(x = X_ML, y = Y, num.trees = 1000, importance = 'impurity', 
+                                classification = TRUE, num.threads = ranger_threads, 
+                                probability = probability, respect.unordered.factors = TRUE)
```
jds485, line 150: @cstillwellusgs: This is how I handled the factor features (PHYSIO in our case)

FlowChar (approved):

PR #147, opened 2022-09-29
added target to plot gage coverage by year

### LINKED ISSUE #133: Plot the time period spanned by each of our gages
Plot the time period spanned by each of our gages to see if we're currently biased to a particular time period
Envisioning a graphic like the coverage plot in the targets-3 training (3_visualize step).

slevin75:
OK, I don't know how to link this to an issue but this pr is for issue # 133.
Closes #133 

I created plots of the years in which each gage has daily streamflow and grouped it by cluster for the mid-high and high quantiles.  We have a lot of gages drop out in the mid 1980s which is odd.

midhighquants_data_coverage

highquants_data_coverage

### DIFF
--- a/3_cluster/src/gage_year_coverage_by_cluster.R
+++ b/3_cluster/src/gage_year_coverage_by_cluster.R
@@ -22,9 +22,12 @@ plot_data_coverage <- function(screened_site_list, cluster_table, dv_data_dir, d
   
   df_tally_allgages<-data.frame()
   for (i in 1:length(screened_site_list)){
-    print(i)
-    
+
     fname<- file.path(dv_data_dir,screened_site_list[i])
+    ##all streamgages in the screened_site_list have 8 digit gage numbers.
+    ##getting the station ID from the file name instead of the downloaded file
+    ##because the site_no column for gage 02490105 has been converted to a
+    ##date and it throws an error when it tries to rbind.
     station.ID <- substr(screened_site_list[i], start = 1, stop = 8)
     df<-read_csv(fname, col_types = cols() ) %>%
       addWaterYear()
@@ -144,6 +147,8 @@ get_estimated_data<- function(fname){
   
   station.ID <- substr(fname, start = 15, stop = 22)
   df<- read_csv(fname, col_types = cols()) 
+  ##returns an empty df for 4 gages where there are more than 1 discharge column in NWIS download.
+  ##these sites are omitted from the plots. 
   if("discharge" %in% names(df)){
     df <- df %>% 
       mutate(ID = station.ID) %>%
--- a/3_cluster/src/gage_year_coverage_by_cluster.R
+++ b/3_cluster/src/gage_year_coverage_by_cluster.R
@@ -1,0 +1,60 @@
+plot_data_coverage<- function(clean_daily_flow, cluster_table, dir_out){
+  
+  #' @description Function to plot the date range for which there is data for each gage. Gages
+  #' are grouped by cluster group.  Will produce a file for each quantile group of clusters.
+  #' 
+  #' @param clean_daily_flow p1_clean_daily_flow target. Includes date, daily discharge
+  #' year, day, and site_no
+  #' @param cluster_table data frame with ID (gage ID), and any number of columns containing
+  #' cluster id for different groups of quantiles
+  #' @param dir_out directory to save plots
+  
+  
+
+  years_with_data<- clean_daily_flow %>%
+    select(site_no, year_val) %>%
+    distinct(.) %>%
+    rename(ID = site_no) %>%
+    left_join(.,cluster_table, by= "ID") 
+  
+  
+  plot_cols <- names(years_with_data[,-c(1:2)])
+  plot_list<- purrr::map(plot_cols, make_por_plot, years_with_data, dir_out)
+  return(plot_list)
+  
+}
+
+
+make_por_plot<- function(plot_col, years_with_data, dir_out){
+  
+  df <- years_with_data  %>%
+    rename(cluster = plot_col) %>%
+    select(ID, year_val, cluster) %>%
+    mutate(cluster_ID = paste0(cluster, "_", ID))
+  
+  cluster_name<- paste(plot_col, "cluster group")
+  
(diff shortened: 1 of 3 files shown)


jds485:
Thanks, Sara! Does this filter out incomplete years of record? Or is this coloring every day that has data? Just trying to think through the drop of gages after the mid-1980s.

slevin75:
This uses the clean_daily_flow target data which only has complete years, so the incomplete years or years with provisional data are filtered out.

jds485:
Okay, thanks. We should discuss revisiting our complete year method on Monday. Not sure how time feasible it is, but it looks like we're missing a lot of data. Would it be quick to make a plot of % data in incomplete years? I'm wondering how close we are to complete records in those years. Can be a separate issue

slevin75:
yeah I think I can get that done.

cstillwellusgs:
Based on my (limited) knowledge of the data review and approval process, I would think estimated data are good for our purposes. The plots suggest to me that the added uncertainty from estimated data is worth the additional data availability. 

No reviewer assigned - is it me?

jds485:
Thanks for doing this analysis! The concern that I have with using estimated data from 1980-present is that it used the MOVE method to estimate from neighboring gages that have records. So we're potentially using correlated information but not considering it in our models (we wouldn't even have a way to know how to consider it without knowing which sites MOVE used to fill in each gap). @FlowChar suggested that the pre-1980 methods to fill in gaps were different, but the method likely would have varied across CONUS. 

Anyway, all that said, the quickest way for us to handle this problem seems to be including the estimated data. I would be interested in how high the flow magnitudes are for those estimated flows, relative to the unestimated flow quantiles. Are we seeing flows > median be filled in? Or is it mostly low flows which for now we are not concerned about? That can be a separate issue to evaluate.

slevin75:
This is ready to go if anyone wants to review it.  Although - does it even need to be merged in? If these are more diagnostic plots rather than results we are going to need, I can just keep it as an open branch.  

I added one last plot to look at which quantile flows were being estimated for each gage.  Histograms are grouped by cluster.  Looks like in most regions, there are some higher flows being estimated.  Region 5(midhigh) and 4 (high) is the rocky mountain region and that seems to have the fewest estimated high flows.   I still think it might be worth including those estimated flows.  There are for sure some pretty substantial trends we are missing without it, especially in the years since 2000.  I could try running a branch using the estimated flows and run it over the weekend and just see how it looks.


image

jds485:
I like this idea of running a branch with the estimated flows included and comparing results. I suggest that you copy your current Tallgrass fork to run these results. That way it will not overwrite the targets without the estimated flows.

3_cluster/src/gage_year_coverage_by_cluster.R
```diff
@@ -0,0 +1,161 @@
...
+  ##now I just removed them - there aren't many and it was breaking my head to try to combine them.
+  
+  df_all <- data.frame()
+  file_list <- list.files(dv_data_dir, pattern = "_dv.csv")
+  screened_site_list <-paste0 (screened_site_list, "_dv.csv") 
+  screened_site_list <- screened_site_list[which(screened_site_list %in% file_list)]
+  
+  df_tally_allgages<-data.frame()
+  for (i in 1:length(screened_site_list)){
+    print(i)
```
jds485, line 25:
```suggestion
```

jds485: or change to message(screened_site_list[i]) to print the site ID

3_cluster/src/gage_year_coverage_by_cluster.R
```diff
@@ -0,0 +1,161 @@
...
+    
+    fname<- file.path(dv_data_dir,screened_site_list[i])
+    station.ID <- substr(screened_site_list[i], start = 1, stop = 8)
```
jds485, line 28: Will this always work? I think some gages are longer than 8 characters

3_cluster/src/gage_year_coverage_by_cluster.R
```diff
@@ -0,0 +1,161 @@
...
+      mutate(ID = station.ID) %>%
+      mutate(quantile = ecdf(discharge)(discharge))
+    
+    df_estimated  <- df %>%
+      filter(., grepl("e", discharge_cd)) %>%
+      select(ID, Date, discharge, discharge_cd, quantile)
+    return(df_estimated)   
+    
+  } #end if
+
```
jds485, line 158: Do you want this to return an empty df when 'discharge' is not in names(df)?

jds485 (approved):
Looks good! I just have some minor comments. I think we can merge this because we'll want to use these plots as supplementary info in the paper

3_cluster/src/gage_year_coverage_by_cluster.R
slevin75: I was worried about that too - but as of now, all our gages have 8 characters.  The reason I had to do this was because there is one gage where the site_no in the csv file has been turned into a date for some reason, so when it gets to that gage, it can't rbind the data to the other sites. So as a work around for now I just extracted the gage number from the file name and replaced the existing site_no column.   I made an issue to look into why that file is messed up. Once that gets fixed, this line could be removed and the existing site_no column could be used.

3_cluster/src/gage_year_coverage_by_cluster.R
slevin75: This was in there because there are 4 files which have more than one discharge column in NWIS for some reason.  Instead of trying to figure it out what to do with them for this pr, I just excluded them.  I figured 4 sites wouldn't make a big difference in the plots.  But - if we want these plots for a paper, and we want to be more precise, I can go back and see how we handled it in the data screening section and get them in there.

3_cluster/src/gage_year_coverage_by_cluster.R
cstillwellusgs: combining those gages is really turning into a pain in the ass, huh?? lol

jds485: Okay, no problem! We can omit those sites from these plots. Can you add comments above the lines with workarounds so we know why it's programmed that way?

PR #152, opened 2022-10-21
Time period plot

slevin75:
OK - here is the pr to make those plots and keep all the estimated data.  Neither of the directories on tallgrass are up to date because I reran the estimated data directory using only the low-estimated data.  I wanted to try to run 1_fetch.R again because initially when I made the change about the estimated data, the pipeline failed because there was a site with negative discharge values, so I added a couple lines to remove any data with negative data,.  But for some reason, I don't see that change in this pr, which is mysterious to me. Anyway, I've been in the tallgrass queue all morning and I'm leaving soon so I won't get to re-run it.  Anyway, if you review it and are able to run it, and it runs ok, then maybe I am just losing my mind. If it crashes, then it probably needs that update for the negative values put back in and I can add it when I get back.

### DIFF
--- a/3_cluster.R
+++ b/3_cluster.R
@@ -979,5 +979,12 @@ p3_targets_list<- list(
                                         rename(midhigh = '0.5,0.55,0.6,0.65,0.7_k5',
                                                high = '0.75,0.8,0.85,0.9,0.95_k5'), 
                                       dv_data_dir = "./1_fetch/out",
-                                      dir_out = '3_cluster/out'))
+                                      dir_out = '3_cluster/out')),
+  tar_target(p3_plot_complete_years,
+             plot_complete_years(clean_daily_flow = p1_clean_daily_flow,
+                                 cluster_table <- p3_gages_clusters_quants_agg_selected %>%
+                                   select(ID, contains('_k5')) %>%
+                                   rename(midhigh = '0.5,0.55,0.6,0.65,0.7_k5',
+                                          high = '0.75,0.8,0.85,0.9,0.95_k5'), 
+                                 dir_out = '3_cluster/out'))
 )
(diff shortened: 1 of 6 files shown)


3_cluster/src/gage_year_coverage_by_cluster.R
```diff
@@ -22,12 +22,9 @@ plot_data_coverage <- function(screened_site_list, cluster_table, dv_data_dir, d
   
   df_tally_allgages<-data.frame()
   for (i in 1:length(screened_site_list)){
-
+    print(i)
```
jds485, line 25:
```suggestion
```

3_cluster.R
```diff
@@ -979,5 +979,12 @@ p3_targets_list<- list(
                                         rename(midhigh = '0.5,0.55,0.6,0.65,0.7_k5',
                                                high = '0.75,0.8,0.85,0.9,0.95_k5'), 
                                       dv_data_dir = "./1_fetch/out",
-                                      dir_out = '3_cluster/out'))
+                                      dir_out = '3_cluster/out')),
+  tar_target(p3_plot_complete_years,
```
jds485, line 983: this should be changed to a file target

3_cluster/src/gage_year_coverage_by_cluster.R
```diff
@@ -163,4 +158,40 @@ get_estimated_data<- function(fname){
...
+          axis.ticks.y = element_blank())+
+    scale_x_continuous(breaks = seq(1900, 2020, 20))+
+    facet_grid(high~., scales="free", space="free")+
+    ggtitle("complete years of data")
+  
+  fname <- paste0("complete_years.png")
+  fileout <- file.path(dir_out, fname)  
+  ggsave(filename = fileout, 
+         plot =   plot_grid(p1,p2, ncol = 1, label_size = 8))
+
```
jds485, line 196:
```suggestion
   return(fileout)
```

jds485 (changes requested):
These changes look good to me. I also have not been able to get on Tallgrass to run this, so I'd say I'm not finished with the review. Let me know when the run you submitted finishes.

Also, because we'll have to re-download all of the NWIS data, we can switch the end_date to 2022-09-30

PR #160, opened 2022-11-09
Event definition

### LINKED ISSUE #137: Add new event definition option based on percentile ranges
Requires edits to how events are defined in EflowStats functions. Recommend adding an argument to compute events only within a specified percentile range.

Use case would be when we aggregate quantiles (50-70, > 70) we can enforce no overlapping events in those aggregated groups

slevin75:
I haven't updated the calcHIT function yet, but in the interest of getting these changes through, I am putting in this pr.  

In this PR, I created find_events_mod function which is the find_events function from EflowStats modified to accept a lower and upper threshold.  The thresholds will be set within in the calcFDC function.  For 'high' threshold types, events are defined as:
	(discharge > threshold_lower) & (discharge <= discharge_upper) 

Within the cal_FDC function, if allow_event_overlap = FALSE , lower thresholds are set from the vector of NE_flows.  This will prevent overlapping events in aggregated groups.  If allow_event_overlap = TRUE, lower threshold stays the same and upper threshold will be max(discharge) for all NE_flows and this will revert back to the original event definition. 
For 'low' threshold types, events are defined as:
       (discharge < threshold_upper + 0.001) & (discharge >= discharge_lower)
0.001 implemented previously in order to get discharge <= threshold_upper for the low threshold type.

Results shown for an example gage.  These plots aren't in the pr, I just coded them up to look at the results really quickly, but I can add them if you want.  

image
image
image
image

Closes #137


jds485 (changes requested):
Thanks, Sara! Your edits look good to me. I have some small comments and suggested edits. I still need to run the code to check that it will work for the use cases I was thinking about.

2_flow_metrics.R
```diff
@@ -101,6 +103,64 @@ p2_targets_list<- list(
...
+                             year_start = year_start,
+                             out_format = 'pivot',
+                             threshold_type = 'low',
+                             allow_event_overlap = FALSE),
+             map(p1_screened_site_list),
+             deployment = 'worker'
+  ),
+  #For these aggregated ranges of events, we would not be interested in the
+  # category above the second NE_quants (70-max in this case). Should remove
+  # that result from these targets.
```
jds485, line 137: @slevin75: I added these 2 example targets that span a larger range than specified in NE_quants. If it looks like we should still aggregate quantiles in the cluster analysis, I'd like to compare cluster results for this method vs. using each individual quantile (as in the above targets). But there may be a better way to do this because we wouldn't need the results for the second NE_quant value. See my comment in the script above this target.

slevin75:
OK, I have updated the find_events_mod function and reran the p2 targets.  I think everything is functioning correctly now.

slevin75:
@jds485 actually - wait do re-review it yet.  I think I messed up one thing. I'm not sure I left in the option to keep overlapping quantiles now.

slevin75:
OK, I think we are good to go now. I addressed the comments above and changed the find_events definiton and ran the p2 targets.  Everything appears to be ok.

2_flow_metrics/src/calc_FDC.R
```diff
@@ -21,7 +22,11 @@ calc_FDCmetrics <- function(site_num, clean_daily_flow, yearType,
   #' @param out_format format for the output table of metrics. The default matches
   #' the EflowStats format. Use 'pivot' for a simpler table.
   #' @param threshold_type one of either 'high' or 'low' indicating if the threshold
-  #' should consider all flows higher or lower than the threshold.
+  #' should consider all flows higher or lower than the threshold.false, events are defined such that
+  #' there are no overlapping events between quantile groups.  If false, the same set of dates may
```
jds485, line 26: Check description. Is the `threshold.false` supposed to be for `allow_event_overlap`?

slurm/tar_make.R
```diff
@@ -3,5 +3,5 @@ arg = commandArgs(T)
 
 source('_targets.R')
 
-tar_make()
+tar_make(starts_with("p2"))
```
jds485, line 6:
```suggestion
tar_make()
```

jds485 (approved):
This looks good to me. And I think you do have the option to retain the overlapping quantiles. There are separate targets.

PR #163, opened 2022-11-14
Misc fixes

### LINKED ISSUE #159: Remove '/' in attribute name: SOHL_HAY/PASTURE_weighted_avg
This causes issues when plotting because the '/' is taken as a directory name.

### LINKED ISSUE #161: Change PHYSIO to factor
Applies to CAT, TOT, ACC attributes

cstillwellusgs:
Closes #159 

Closes #161


jds485 (approved):
That looks good to me. I don't think you need to re-run all targets because that would take a while to update. I can address updating some targets when I run the multiclass PR targets.

PR #171, opened 2022-12-04
feature attributes for conus

### LINKED ISSUE #142: Gathering watershed attributes NHD reaches across CONUS
We currently download all NHD data and save only the reaches with gages. We now want to save attributes for all reaches.

Suggested method:
1. Make a new target that downloads the data and saves it, branching over the individual datasets (e.g., climate dataset, soils dataset)
2. Edit the existing target to select the gages reaches from the target in 1.

I don't think we'll be able to break the download into smaller sizes (e.g., based on HUC region). But if each CONUS dataset is too large in file size, we could store the final dataset by HUC.

### LINKED ISSUE #153: Fetch CONUS reaches and flags
Add target to `1_fetch.R` to fetch NHD reaches across CONUS

Also fetch flags like tidal indicators that identify reaches for which we would not want to make a prediction.

cstillwellusgs:
closes #142 
closes #153


cstillwellusgs:
@jds485 `p1_feature_vars_g2' and `p1_feature_vars_conus` do not have an identical number of columns. `p1_feature_vars_g2` contains some random columns from gages2.1 (npdes, strg, etc.) and also the weighted averages for time-varying attributes (land cover, weather, dams). Other than those differences, the remaining columns all match (but they are not in the same order). Do any of your downstream functions rely on specific positioning and/or dimensions of the `p1_feature_vars_g2` data frame (and thus the specific positioning and/or dimensions of the `p1_feature_vars_conus` data frame)? If I remember correctly, the function that down-selects correlated variables relies on the order.  

Down the road I intend to clean up the `prep_feature_vars_g2()` function and at that point I can change the order and inconsistency of variables across `p1_feature_vars_g2` and `p1_feature_vars_conus` but for the sake of time I'm assuming the issue I described above is okay for now. Let me know if I'm assuming incorrectly.

cstillwellusgs:
it also appears that I have a merge conflict but the "Resolve Conflict" button is greyed out... maybe we can scan that together in the morning if you don't mind.

jds485:
> Do any of your downstream functions rely on specific positioning and/or dimensions of the p1_feature_vars_g2 data frame?

Yes for `p1_feature_vars_g2`, but no for `p1_feature_vars_conus` because we can select the columns to be the same as those selected after processing `p1_feature_vars_g2`. So I don't see column order as a problem

.gitignore
```diff
@@ -19,6 +19,9 @@ rstudio*
 #Gages data
 Gages2.1_RefSiteList.xlsx
 
+#EcoFlows data
+6_predict/in/EcoFlowsAttrs.csv
+
```
jds485, line 24: I committed this file in our repo, so should be deleted from .gitignore
```suggestion
```

1_fetch/src/get_feature_vars.R
```diff
@@ -1,5 +1,52 @@
+finalize_vars_conus <- function(prep1_conus, prep2_conus, drop_attr) {
+  
+  #'@description combines outputs from prep1 and prep2 conus functions to provide
+  #'full set of feature variables for all comids in conus
+  #'
+  #'@param prep1_conus filepaths for all variables except land cover
```
jds485, line 6:
```suggestion
  #'@param prep1_conus filepaths for all variables except
```

1_fetch/src/get_feature_vars.R
```diff
@@ -1,5 +1,52 @@
+finalize_vars_conus <- function(prep1_conus, prep2_conus, drop_attr) {
+  
+  #'@description combines outputs from prep1 and prep2 conus functions to provide
+  #'full set of feature variables for all comids in conus
+  #'
+  #'@param prep1_conus filepaths for all variables except land cover 
+  #'land cover variables (from prep1_vars_conus)
+  #'@param prep2_conus filepath for land cover variables (from prep2_vars_conus)
+  #'@param retain_attr character list of COMID and other attributes used for 
+  #'left joins
```
jds485, line 10:
```suggestion
  #'@param drop_attr character vector of attributes to drop from
  #' prep1_conus tables before joining
```

1_fetch/src/get_feature_vars.R
```diff
@@ -1,5 +1,52 @@
...
+  #'@param outdir filepath for final data downloads
+  #'@param seven_zip filepath (full, not relative) to 7zip executable
+  #'
+  #'@return unzipped nhd geodatabase
+  
+  download_nhdplusv2(
+    outdir,
+    url = paste0("https://edap-ow-data-commons.s3.amazonaws.com/NHDPlusV21/",
+                 "Data/NationalData/NHDPlusV21_NationalData_Seamless", 
+                 "_Geodatabase_Lower48_07.7z"),
```
jds485, line 37: Is it possible to specify a specific version of this file to grab from S3? Do you know if this link will always grab the most recent?

1_fetch/src/get_feature_vars.R
```diff
@@ -198,9 +268,9 @@ prep_feature_vars <- function(sb_var_data, sites_all, sites_screened,
                                  class == 7 ~ "BARREN", class == 8 ~ "FOREST", 
                                  class == 9 ~ "FOREST", class == 10 ~ "FOREST", 
                                  class == 11 ~ "GRASSLAND", class == 12 ~ "SHRUBLAND", 
-                                 class == 13 ~ "CROPLAND", class == 14 ~ "HAY/PASTURE", 
+                                 class == 13 ~ "CROPLAND", class == 14 ~ "HAYPASTURE",
```
jds485, line 271: In [this issue](https://github.com/USGS-R/regional-hydrologic-forcings-ml/pull/163/files) we convert these from `/` to `-`.
```suggestion
                                 class == 13 ~ "CROPLAND", class == 14 ~ "HAY-PASTURE", 
```

1_fetch/src/get_feature_vars.R
```diff
@@ -198,9 +268,9 @@ prep_feature_vars <- function(sb_var_data, sites_all, sites_screened,
                                  class == 7 ~ "BARREN", class == 8 ~ "FOREST", 
                                  class == 9 ~ "FOREST", class == 10 ~ "FOREST", 
                                  class == 11 ~ "GRASSLAND", class == 12 ~ "SHRUBLAND", 
-                                 class == 13 ~ "CROPLAND", class == 14 ~ "HAY/PASTURE", 
+                                 class == 13 ~ "CROPLAND", class == 14 ~ "HAYPASTURE", 
                                  class == 15 ~ "WETLAND", class == 16 ~ "WETLAND", 
-                                 class == 17 ~ "ICE/SNOW")) %>%
+                                 class == 17 ~ "ICESNOW")) %>%
```
jds485, line 273: In [this issue](https://github.com/USGS-R/regional-hydrologic-forcings-ml/pull/163/files) we convert these from `/` to `-`.
```suggestion
                                 class == 17 ~ "ICE-SNOW")) %>%
```

1_fetch/src/get_feature_vars.R
```diff
@@ -399,7 +469,7 @@ prep_feature_vars <- function(sb_var_data, sites_all, sites_screened,
            year = str_sub(name, 14, 17)) %>%
     group_by(COMID, unit) %>%
     summarise(avg_annual = mean(value, na.rm = TRUE), .groups = "drop") %>%
-    mutate(label = paste0(unit, "_AVG_WILDFIRE_shorterm_avg")) %>%
+    mutate(label = paste0(unit, "_WILDFIRE_shortterm_avg")) %>%
```
jds485, line 472: I think this will cause downstream modeling targets to be outdated. For now, I think we should keep as we had it before.
```suggestion
    mutate(label = paste0(unit, "_AVG_WILDFIRE_shorterm_avg")) %>%
```

1_fetch/src/get_feature_vars.R
```diff
@@ -489,3 +559,246 @@ prep_feature_vars <- function(sb_var_data, sites_all, sites_screened,
...
+  #'
+  #'@param sb_var_data target generated from the 'get_sb_data()' function mapped over the 
+  #''p1_sb_var_ids' targets; the 'p1_sb_var_ids' target reads in a .csv of ScienceBase
+  #'identifiers and names, and the 'get_sb_data()' function generates a list of .csv file
+  #'locations containing the feature variable data from each ScienceBase identifier
+  #'joined with the COMIDs contained in the 'sites' parameter
+  #'@param sites data frame with a COMID column including all reaches of interest
+  #'@param retain_attr character strings of additional column headers (attributes) 
+  #'to keep in the sites data frame
+  #'@param out_file_label naming convention to label output files from this function
```
jds485, line 576:
```suggestion
  #'@param out_dir output file directory
  #'@param out_file_label naming convention to label output files from this function
```

1_fetch/src/get_feature_vars.R
```diff
@@ -489,3 +559,246 @@ prep_feature_vars <- function(sb_var_data, sites_all, sites_screened,
...
+  write_csv(data, file_out)
+  return(file_out)
+}
+
+
+prep2_vars_conus <- function(sohl_early, sohl_late, outdir, out_file_label) {
+  
+  #'@description joins all data downloaded from ScienceBase with sites of interest
+  #'
+  #'@param prep1_vars target generated from the 'prep1_vars_conus()' function
```
jds485, line 723: should be `sohl_early` and `sohl_late` instead

1_fetch/src/get_feature_vars.R
```diff
@@ -489,3 +559,246 @@ prep_feature_vars <- function(sb_var_data, sites_all, sites_screened,
...
+                    show_col_types = FALSE)
+  late <- read_csv(paste0(outdir, "/sb_", sohl_late, "_raw_conus.csv"), 
+                   show_col_types = FALSE)
+  late <- late %>%
+    select(COMID, contains("SOHL00"))
+  sohl <- left_join(early, late, by = "COMID")
+  sohl[sohl == -9999] <- NA_real_
+  
+  unit <- c("CAT", "ACC", "TOT")
+  category <- as.character(1:17)
```
jds485, line 740: Can this be generalized instead of hard coded?

1_fetch/src/get_feature_vars.R
```diff
@@ -489,3 +559,246 @@ prep_feature_vars <- function(sb_var_data, sites_all, sites_screened,
...
+                                na.rm = TRUE),
+             MINING = SOHL6, 
+             BARREN = SOHL7, 
+             GRASSLAND = SOHL11, 
+             SHRUBLAND = SOHL12, 
+             CROPLAND = SOHL13, 
+             HAYPASTURE = SOHL14, 
+             WETLAND = rowSums(select(., SOHL15, SOHL16),
+                                 na.rm = TRUE), 
+             ICESNOW = SOHL17) %>%
```
jds485, line 770: In this function, do `HAYPASTURE` and `ICESNOW` need to be edited here to have `-`? Does the g2 processing function also need this edit?

jds485 (changes requested):
Look good! I have some function arg. description edits. See my comments about changing `/` to `-` and about S3 file version.

1_fetch/src/get_feature_vars.R
cstillwellusgs: @ajsekell do you know the answer to Jared's question?

1_fetch/src/get_feature_vars.R
cstillwellusgs: Yes but sometimes the "-" in a column header can cause issues with dplyr functions (reads it as a minus sign) so I chose to delete completely (updated for the g2 and conus feature variable data frames).

1_fetch/src/get_feature_vars.R
cstillwellusgs: yeah you are right... so I guess I will revert the HAYPASTURE to Hay_PASTURE and ICESNOW to ICE_SNOW as well

jds485: You can file an issue to rename these columns. That can be done with the reorganization of the column order

1_fetch/src/get_feature_vars.R
jds485: Oh, okay. I didn't know that would happen with character names

cstillwellusgs:
All comments addressed. Did you see Andrew's response about the nhdPlusTools url? Other than that, it should be good to go.

1_fetch/src/get_feature_vars.R
```diff
@@ -112,11 +159,34 @@ get_sb_data_log <- function(sb_var_ids, file_out) {
...
+  #'@return non-tidal conus-wide comids of specified ftype(s)
+  #'with specified attributes retained from geodatabase (class 'sf')
   
+  nhd_full <- st_read(nhd_conus_gdb, layer = 'NHDFlowline_Network')
+  comid_conus <- nhd_full %>%
+    select(all_of(attrib_to_keep)) %>%
+    filter(FTYPE %in% all_of(ftype_to_keep)) %>%
+    filter(Tidal == 0)
+  
+  return(comid_conus)
```
jds485, line 184: let's transform to EPSG 4326 coordinate system before returning this file. That's what we use for g2 data. May also need to remove the ZM with st_zm().

PR #174, opened 2022-12-06
Maps for CONUS reaches

### LINKED ISSUE #149: Create map of predicted regions across CONUS
Depends on #144 model for cluster regions, #153 for fetching reaches, and #142 for gathering CONUS reach data.

Function created within #136 to generate maps for highest to lowest probability region, one map per region. Need to create targets for:
- predicting reach class probabilities
- plotting those probabilities as maps

jds485:
This PR provides targets for region predictions for CONUS reaches, and makes maps for CONUS reaches. Several functions have been edited to be parallelized.

This PR also provides plot aesthetic edits.
- Note that the whole `seasonal_metric_cluster.R` script shows as outdated, but only 1 line of `plot_cluster_map` function changed to increase the font size: `strip.text.x = element_text(size = 20)`

Closes #149


slevin75 (approved):
OK, this looks good to me.  Those NHD maps are really nice looking!

PR #184, opened 2022-12-20
Remove tidal from training

### LINKED ISSUE #172: Check select(-contains()) for unintended attributes
Example: `refine_features` function drops longterm averages with 'LON' specified. That's okay because the weighted averages were retained, but would be better to specify all columns that are dropped.

### LINKED ISSUE #173: Remove tidal gages from training
Use tidal flag from `p1_sites_conus` to remove the tidal gages from the gagesii sites.

cstillwellusgs:
Removed tidal gages (per nhdPlus) from training dataset in `prep_feature_vars_g2()`. Also, checked for any issues with `contains()` or `starts_with()` and didn't find anything. 

closes #173 
closes #172

### DIFF
--- a/1_fetch.R
+++ b/1_fetch.R
@@ -227,26 +227,6 @@ p1_targets_list <- list(
              format = "file"
   ),
   
-  ##merge and select feature variables from gagesii list
-  tar_target(p1_feature_vars_g2, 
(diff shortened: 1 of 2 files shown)


jds485 (approved):
Thanks! Looks good to merge

PR #186, opened 2022-12-21
Add annual flow metric clusters

### LINKED ISSUE #139: Add cluster analysis based on annual flow metric values
Cluster analysis based on annual flow metric values (frequency, duration, volume, peak flow) 
- individual quantiles
- aggregated quantiles

These results will be compared to clustering based on seasonal timing (which we have already done)

jds485:
Draft - target building in progress

This PR adds targets that create clusters based on annual flow metric values. `5_EDA.R` shows as completely different. I think this is a result of line endings being different from this commit to the previous. I added 1 new target to `5_EDA.R` to plot flow metrics for each of the clusters that were discovered using raw flow metric values.

Also updated:
- calc_fhfdc_metrics function: this was slow for computing seasonal flow metric values. I changed the function to be significantly faster (about 8 times faster). I changed `dplyr::do` to `dplyr::summarize`, which contributed to some of the speed up. Most of the speed up resulted from using the output of another function as input to this function when seasonal metrics are requested.
- tar_make SLURM and R scripts: R script edited to make more efficient use of clustermq parallelization (p3 and p4 targets) while using tar_make for all other targets (with built-in parallelization). SLURM script increases the default number of cores and RAM to request. These can still be overwritten with `sbatch` command line flags

Closes #139


slevin75:
@jds485  Code looks fine to me.  I didn't look at the targets since they may still be running but I think the code changes are ok to merge.

jds485:
The moving window targets are still not finished building. I think they're close to finishing so I'll restart building them again today. But after they're done I'm going to comment the targets out again.

PR #191, opened 2023-02-07
added PDP panel plot function and target for paper

### LINKED ISSUE #190: Panel plot for top features - multiclass model
Create a 6 panel plot for:
CAT PPT APR
CAT PPT JUN
ACC PPT AUG
ACC WB FEB
ACC TAV FEB
ACC LSTFZ

Compared to current plots:
- Change the theme to have a grid
- Enlarge axis labels
- Trim y-axis range

jds485:
This PR adds a function to make a panel plot of PDPs. The function has several lines that are specific to the attributes that we're plotting in the paper figure. I'm not sure of a good way to generalize, but I'm also not sure we'd need to have it general.

Result:
PDP_offset_panel_RF_multiclass_high_NoPhysio_6panel

Noting that running tar_make() to create this plot provides different plot dimensions and sizes:
PDP_offset_panel_RF_multiclass_high_NoPhysio

Something that might be good for the paper is to add a rug to these plots to indicate where the data support are located. I do not know how to get the geom_point() to plot and not show in the legend.

Closes #190

### DIFF
--- a/6_predict/src/plot_diagnostics.R
+++ b/6_predict/src/plot_diagnostics.R
@@ -881,18 +881,19 @@ plot_pdp_panel <- function(partial, data, model_name, out_dir, ice = FALSE,
   p <- ggplot(data = data_facet, 
               mapping = aes(x = variable, y = yhat, color = yhat.id)) + 
     geom_line(size = 1) + 
+    #geom_point(mapping = aes(x = variable, y = -0.3), show.legend = FALSE, inherit.aes = FALSE) +
     ylim(plt_ylim) +
     ylab(plt_ylab) +
     xlab('Attribute Value') +
     theme_bw() +
-    theme(strip.text = element_text(size = 6), axis.text = element_text(size = 6))+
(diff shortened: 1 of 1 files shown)


cstillwellusgs:
Try this out, there is an argument for `show.legend = FALSE` that could probably work.

cstillwellusgs:
To prevent `tar_make()` from plotting different dimensions, you canprobably specify all font sizes, margin widths, etc. in the `theme()` and then overall dimensions in `ggsave()`. That probably isn't necessary for now but we can address it once we need a publication-ready proof from the journal.

jds485:
> Try this out, there is an argument for `show.legend = FALSE` that could probably work.

Thanks! That does work, but for now I'm going to omit this edit because I realized that each region could have a different range of values for each attribute and that will take some time to work up into the plot. This is the function call that worked for me. The x `variable` is not the data location and is instead the 25 grid points at which the partial dependence was computed (evenly spaced along each x-axis).

`geom_point(mapping = aes(x = variable, y = -0.3), show.legend = FALSE, inherit.aes = FALSE)`

PR #195, opened 2023-02-28
Add function and target to create map of transition regions.

### LINKED ISSUE #194: Create plots for "transition" regions
- Plot transition zones - case_when() likely_rank_1 >= "0.5" & likely_rank_2 >= "0.25 (for instance, play with values) 
- Consider 20 cases (directiona important) and 10 cases (direction unimportant)
- May be less than 20/10 cases with substantial number of reaches (not worth mapping all cases)

cstillwellusgs:
Closes #194 

All reaches that have 2nd most likely probability greater than 0.1
Transitions_High_NoPhysio_NHD_fine_k5_threshold_0 1

All reaches that have 2nd most likely probability greater than 0.2
Transitions_High_NoPhysio_NHD_fine_k5_threshold_0 2

### DIFF
--- a/6_predict/src/plot_diagnostics.R
+++ b/6_predict/src/plot_diagnostics.R
@@ -572,11 +572,58 @@ make_transition_region_map <- function(class_probs, reaches, threshold,
     geom_polygon(fill = "white", color = "gray80") +
     geom_sf(data = reaches_to_plot, mapping = aes(color = plot_group), 
             inherit.aes = FALSE, size = pt_size) +
-    facet_wrap(. ~ two_mixed, ncol = 2) +
+    facet_wrap(. ~ two_mixed, ncol = 4) +
     scale_color_manual(values = c("#1b9e77", "#7570b3")) +
     labs(x = "Longitude", y = "Latitude", color = "") +
     theme(legend.position = "bottom") +
-    guides(color = guide_legend(override.aes = list(size=2)))
+    guides(color = guide_legend(override.aes = list(size = 1)))
+  fname = paste0(out_dir, model_name, "_threshold_", threshold, ".png")
+  ggsave(filename = fname, bg = "white",
+         height = 7.5, width = 10, units = "in", dpi = 300)
+  
+  return(fname)
+}
+
+
+make_region_count_map <- function(class_probs, reaches, threshold,
+                                  out_dir, model_name, pt_size = 0.5) {
+  #' @description this function creates a map showing the reaches colored by 
+  #' the number of clusters (regions) predicted above a certain probability
+  #' 
+  #' @param class_probs dataframe of predicted class probabilities for each reach.
+  #' must have an "ID" column and columns (ordered 1 through 5) for the class 
+  #' probabilities labeled with the name of the class. No other columns. 
+  #' @param reaches sf object containing the reaches to plot. Must have a "ID" column
+  #' corresponding to the same "ID" values in class_probs
+  #' @param out_dir where output figures are saved
+  #' @param threshold minimum probability for region to be counted "probable"
+  #' @param model_name name to add to the file name that describes this model
+  #' @param pt_size size of features to be mapped
+  #' 
+  #' @return file paths to maps
+  
+  #identify region rankings based on probability
+  region_prob_count <- select(class_probs, -ID) %>%
+    rename(prob_1 = 1, prob_2 = 2, prob_3 = 3, prob_4 = 4, prob_5 = 5) %>%
+    mutate(num_prob_thresh = rowSums(. > threshold))
+  region_prob_count <- 
+    bind_cols(tibble(ID = class_probs$ID), region_prob_count)
+  
+  #Join ranks to reaches for plotting
+  reach_prob_count <- left_join(reaches, region_prob_count, by = "ID")
+  
+  #Plot map with colors counting the number of "probable" regions
+  states <- map_data("state")
+  region_count_map <- ggplot(states, aes(x = long, y = lat, group = group)) +
+    geom_polygon(fill = "white", color = "gray80") +
+    geom_sf(data = reach_prob_count, mapping = aes(color = num_prob_thresh), 
+            inherit.aes = FALSE, size = pt_size) +
+    scale_color_viridis() +
+    labs(x = "Longitude", y = "Latitude", 
+         color = paste0("Number of seasonal regions with probability ", 
+                        "greater than ", threshold)) +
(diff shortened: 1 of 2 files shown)


6_predict/src/plot_diagnostics.R
```diff
@@ -447,6 +447,101 @@ make_class_prediction_map <- function(class_probs, reaches, out_dir,
   return(fnames)
 }
 
+
+make_transition_region_map <- function(class_probs, reaches,
+                                       region_threshold,
+                                       out_dir, model_name, pt_size = 0.5) {
+  #' @description this function creates a map showing the transition regions
```
jds485, line 454:
```suggestion
  #' @description this function creates a panel plot of maps showing the transition regions 
```

6_predict/src/plot_diagnostics.R
```diff
@@ -447,6 +447,101 @@ make_class_prediction_map <- function(class_probs, reaches, out_dir,
...
+make_transition_region_map <- function(class_probs, reaches,
+                                       region_threshold,
+                                       out_dir, model_name, pt_size = 0.5) {
+  #' @description this function creates a map showing the transition regions 
+  #' (reaches not distinctly predicted in single cluster)
+  #' 
+  #' @param class_probs dataframe of predicted class probabilities for each reach.
+  #' must have an "ID" column and columns for the class probabilities labeled with
+  #' the name of the class. No other columns. 
+  #' @param reaches sf object containing the reaches to plot. Must have a "COMID" column
```
jds485, line 460: COMID is set to ID in the function call. 
```suggestion
  #' @param reaches sf object containing the reaches to plot. Must have an "ID" column corresponding to the same "ID" values in class_probs
```

6_predict/src/plot_diagnostics.R
```diff
@@ -447,6 +447,101 @@ make_class_prediction_map <- function(class_probs, reaches, out_dir,
...
+  #' @param out_dir where output figures are saved
+  #' @param region_threshold minimum probability for region to be considered transitional
+  #' @param model_name name to add to the file name that describes this model
+  #' @param ncores number of cores to use
```
jds485, line 464:
```suggestion
```

6_predict/src/plot_diagnostics.R
```diff
@@ -447,6 +447,101 @@ make_class_prediction_map <- function(class_probs, reaches, out_dir,
...
+  #' @param pt_size size of features to be mapped
+  #' 
+  #' @return file paths to maps
+  
+  #identify region rankings based on probability
```
jds485, line 469:
```suggestion
```

6_predict/src/plot_diagnostics.R
```diff
@@ -447,6 +447,101 @@ make_class_prediction_map <- function(class_probs, reaches, out_dir,
...
 }
 
+
+make_transition_region_map <- function(class_probs, reaches,
+                                       region_threshold,
+                                       out_dir, model_name, pt_size = 0.5) {
+  #' @description this function creates a map showing the transition regions 
+  #' (reaches not distinctly predicted in single cluster)
+  #' 
+  #' @param class_probs dataframe of predicted class probabilities for each reach.
```
jds485, line 457: This also expects that the columns are ordered 1, 2, ... 5 for the regions, right?

6_predict/src/plot_diagnostics.R
```diff
@@ -447,6 +447,101 @@ make_class_prediction_map <- function(class_probs, reaches, out_dir,
...
+  #' @param ncores number of cores to use
+  #' @param pt_size size of features to be mapped
+  #' 
+  #' @return file paths to maps
+  
+  #identify region rankings based on probability
+  #identify region rankings based on probability
+  region_probs <- select(class_probs, -ID) %>%
+    rename(prob_1 = 1, prob_2 = 2, prob_3 = 3, prob_4 = 4, prob_5 = 5)
+  region_ranks <- t(apply(-region_probs, 1, rank, ties.method = "random")) %>%
```
jds485, line 473: I made a `decreasing_rank` [function](https://github.com/USGS-R/regional-hydrologic-forcings-ml/blob/2c84ee34203473f3a0893e7b9ef8a320f87457ec/6_predict/src/plot_diagnostics.R#L471) that could be used here instead of rank of the negative values 

Also, if this takes a long time, could run with parApply instead. I used that for the other CONUS NHD maps

jds485 (changes requested):
These look good! I have some suggestions:
I like using 0.1 map 10% chance is high to me. 

Try adding color to reaches that have > 5% for 3-5 most likely (maybe in gray). Could use all(probs > 0.05) to do this filter.

I'm thinking about labels and layout for the paper. This might be better as a new target/function for the paper graphic:
-	1&2 next to 2&4. Label as Region #&#: rain snow transitions
-	1&5 next to 4&5. Label as Region #&#: rain snow transitions
-	3&5 next to 2&5. Label as Region #&#: different rainfall processes
-	1&4. Label as Region #&#: snowpack transitions
-	2&3. Label as Region #&#: aridity transitions
-	Do not need 3&4 or 1&3 in final graphic (very few sites)

cstillwellusgs:
> Try adding color to reaches that have > 5% for 3-5 most likely (maybe in gray). Could use all(probs > 0.05) to do this filter.

Would this be a separate panel? Or overlaid? I started to think that through but it doesn't really go with the current format.

jds485:
> Would this be a separate panel? Or overlaid

I was thinking overlaid on each panel

cstillwellusgs:
> > Would this be a separate panel? Or overlaid
> 
> I was thinking overlaid on each panel for only the reaches in that panel

Okay, so if any reach has 3 regions with 3+ probabilities of 5% or greater... that reach would be classified completely separate from the transition regions. For example, if a reach is 55% region 1, 30% region 2, and 15% region 3, that would be classified exclusively ("confused" - I'll think of a better label) from the transition regions 1-2, 2-3, and 1-3. Do I understand correctly?

cstillwellusgs:
Transitions_High_NoPhysio_NHD_fine_k5_threshold_0 05

6_predict/src/plot_diagnostics.R
jds485: @cstillwellusgs - checking if this is true

jds485 (approved):
Looks good! I like the color schemes. There's one minor comment to check from the last review, but it looks ready to merge.

